PR #5, opened 2019-10-29
Project 6 - Weather App Bollnas

nyblad:
Live site: https://project-6-weatherapp.netlify.com/

- I went for the 5 days/3 hours forecast and shows the forecast every 6th hour (skipping every other in the array)
- Background image changes before/after sunset and depending on screen size (but have to reload page if you have it opened when passing the sunset/sunrise or changing screen size - something I can fix?)
- The accordion in desktop is not the best, would rather want to show the accordion downwards than to the right
- Feel a bit insecure in how I should structure the js-code, where to put the consts and so on, would appreciate som tips on that
- Made a new fork, therefore no earlier commits (you can see some more in [my Trellocard](https://trello.com/c/tojEiKeH/103-week-6-project) for the whole project process)

/Sofie


code/index.html
```diff
@@ -0,0 +1,59 @@
...
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <link rel="stylesheet" href="style.css">
+  <link href="https://fonts.googleapis.com/css?family=Hind+Guntur&display=swap" rel="stylesheet">
+  <title>Project 6: Weather app</title>
+  <!-- FAVICONKIT -->
+  <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png">
+  <link rel="icon" type="image/png" sizes="32x32" href="/favicon-32x32.png">
+  <link rel="icon" type="image/png" sizes="16x16" href="/favicon-16x16.png">
```
ericaglimsholt: Really nice of you thinking about adding a favicon! Here's a design tip. Now when Chrome has dark mode as a setting the black favicons doesn't show up as you really want. So now you have to thing about adding a white background, or make a script where you fetch if it's dark mode or not (never done this but probably not so complicated)

image

code/script.js
```diff
@@ -0,0 +1,180 @@
...
+    const sunrise = new Date(json.sys.sunrise * 1000)
+    const sunset = new Date(json.sys.sunset * 1000)
+
+    //Declare new variable to show only HH:MM
+    const sunriseTime = sunrise.toLocaleTimeString([], { timeStyle: 'short' })
+    const sunsetTime = sunset.toLocaleTimeString([], { timeStyle: 'short' })
+
+    //THE WEATHER SHOWN IN MAIN-TODAY
+    theCity.innerHTML = `<h2>${json.name}</h2>`
+    theTemp.innerHTML = `<h1>${json.main.temp.toFixed(1)} &deg;C</h1>`
```
ericaglimsholt: Hm, I don't really understand why this suits as a `<h1>` element, but please enlighten me why!

code/script.js
```diff
@@ -0,0 +1,180 @@
...
+    const currentTime = new Date().toLocaleTimeString([], { timeStyle: 'short' })
+
+    let dayTime
+    if (currentTime > sunriseTime && currentTime < sunsetTime) {
+      dayTime = true
+    } else {
+      dayTime = false
+    }
+
+    if (dayTime === true && window.matchMedia("(max-width: 750px)").matches) {
```
ericaglimsholt: Because `dayTime` is a boolean (true or false) you don't need `dayTime === true`, you can have it like `dayTime` only. It works the same but is easier to read! It says "if `dayTime` is true do this"

code/script.js
```diff
@@ -0,0 +1,180 @@
...
+      mainTop.style.backgroundImage = "url('assets/camp-day-small.jpg')"
+    } else if (dayTime === true && window.matchMedia("(min-width: 750px)").matches) {
+      mainTop.style.backgroundImage = "url('assets/camp-day-big.jpg')"
+    } else if (dayTime === false && window.matchMedia("(max-width: 750px)").matches) {
```
ericaglimsholt: And here you can have it like `!dayTime`. That stands for `if `dayTime` is not true`. `!` means "not"

code/script.js
```diff
@@ -0,0 +1,180 @@
...
+        <div class="forecast-time">${shortTime}</div>
+        <div class="forecast-icon">${weatherIcon}</div>
+        <div class="forecast-temp">${averageTemp.toFixed(1)} &deg;C</div>
+        </section>
+        
+        <section class="forecast-details">
+        <p>Weather: ${forecast.weather[0].description}</p>
+        <p>Wind: ${forecast.wind.speed.toFixed(0)} m/s</p>
+        <p>Max temp: ${forecast.main.temp_max.toFixed(1)} &deg;C</p>
+        <p>Min temp: ${forecast.main.temp_min.toFixed(1)} &deg;C</p>
```
ericaglimsholt: You can indent these lines to make it easier to read

code/script.js
nyblad: Np reason for H1 or H2 at all, removed it and updated CSS!

code/script.js
nyblad: Thanks for tip - I have changed it now

nyblad:
Thanks for all the tips, updated the section with icons if-statements and think I got it more readable now. Went for the ID:s instead and made a const instead of typing it on every row:

    const weatherId = json.weather[0].id
    let weatherIcon

ericaglimsholt:
> Thanks for all the tips, updated the section with icons if-statements and think I got it more readable now. Went for the ID:s instead and made a const instead of typing it on every row:
> 
> ```
> const weatherId = json.weather[0].id
> let weatherIcon
> ```
> 
> The background seems to work fine when changing before/after susnet now, must have been cached in my browser or something.
> 
> Just realized how to set a wrapper to make the accordion go downwards in desktop!

Nice improvements! 👏🏻

PR #7, opened 2019-11-01
Project 6 - Jola Chabik

ucha3i:
I'm pretty much done with the app. It was really interesting and cool assignment.

https://ucha3i-project-6.netlify.com/

I'm still working with the list - I'd like to make it look a little nicer, also I'd like to show the names of days  instead of dates in the forecast part.


script.js
```diff
@@ -0,0 +1,189 @@
...
+        /*  if (json.weather === "clouds") {
+             container.innerHTML = `<h1>The weather in ${city} is looking cloudy today!`;
+         } else {
+             container.innerHTML = `<h1>Get your sunnies ready! ${city} is looking for great weather today!</h2>`
+         } */
+
+
+        //IF STATEMENT BASED ON WEATHER ID = json.weather[0].id
+
+        const id = json.weather[0].id
```
ericaglimsholt: What is this an id for? Maybe explain that in the name

script.js
```diff
@@ -0,0 +1,189 @@
...
+            container.innerHTML = `<h1>Get your beanie ready! ${city} is looking for snow today!</h2>`
+        } else if (id >= 701 && id <= 781) {
+            container.innerHTML = `<h1>Missing London? ${city} is looking a foggy weather today!</h2>`
+        } else if (id === 800) {
+            /*  if (nightTime) { */
+            /* container.innerHTML = `<h1>When was the last time you watch the stars? ${city} is good place to find some tonight!</h2>`
+            } else { */
+            container.innerHTML = `<h1>Get your sunnies ready! ${city} is looking for a lot of sun today!</h2>`
+        } else if (id >= 801 && id <= 804) {
+            container.innerHTML = `<h1>Don't forget your favourite hoodie today! Could be a little cloudy in ${city}!</h2>`
```
ericaglimsholt: I don't think this messages suits as `<h1>` element. An `<h1>` should explain the page like "Weather in Lisabon". Maybe this will suit better as a `<h2>` element?

script.js
```diff
@@ -0,0 +1,189 @@
...
+            } else { */
+            weatherImage.src = "glasses.png"
+        } else if (id >= 801 && id <= 804) {
+            weatherImage.src = "cloud.png"
+        }
+
+        /* .catch((err) => {
+                console.log('caught errors', err)
+            }
+        } */
```
ericaglimsholt: Why is this part out commented?

script.js
ericaglimsholt: Wait, when I take a second look at this it starts with a `<h1>` end ends with a `</h2>`?

script.js
ucha3i: I name it id cuz I based my if on weather ID from the website. I'll change that. :)

script.js
ucha3i: Wow, that's messy! You're totally right, I'm sorry. I struggle with that if-statement a lot, and apparently I was too happy that it actually works to take a closer look on that. I'll fix it.

script.js
ucha3i: I wasn't sure if I should leave it few times in a code, I mean for every API.

PR #15, opened 2019-11-03
Weather App project 6 Angelica B

Angelicabrodin:
A link to my site on netlify:
angiesweatherapp.netlify.com

I really enjoyed and struggled with this project :)
I thought it was tricky to understand exactly what was the objects and what was the arrays and also what index nr and so on.. but I managed to fetch the information that we were supposed to get, and I was even able to retrieve the logo from the site that tells the actual weather situation.

Something that I put several hours on, is that I wanted to use my chosen pictures instead of the ones from the webpage, so I tried to do two diffrent functions to do this. But failed, but I have put both of the functions that I tried to do in my js. file so I really would appreciate if you could give me any feedback on how to do it. they are commented out at the bottom. 

All the best, 
A


Angelicabrodin:
Hey Erica, I tried it there now, and for me it is working, hmm.. that is weird?
It took some time for it to load, maybe that is why?

ericaglimsholt:
> Hey Erica, I tried it there now, and for me it is working, hmm.. that is weird?
> It took some time for it to load, maybe that is why?

Something was wrong from Netlifys end! So no worries :)

style.css
```diff
@@ -0,0 +1,205 @@
...
+justify-content: center;
+align-items: center;
+background-color:lightblue;
+height:100vh;
+max-width: 667px;
+border:2px solid;
+border-radius: 16px;
+background-image: url(weather.jpeg);
+background-size: cover;
+background-repeat: no-repeat;
```
ericaglimsholt: Indent one step

weather.js
```diff
@@ -0,0 +1,118 @@
...
+        currentIcon.innerHTML = `<img src"rain.png" alt="Rainy"/>`
+    } else if (currentWeather === "thunderstorm") {
+        currentIcon.innerHTML = `<img src"thunder.png" alt="Thunder"/>`
+    } else if (currentWeather === "	snow") {
+        currentIcon.innerHTML = `<img src"snow.png" alt="Snowing"/>`
+    } else {
+        currentIcon.innerHTML = `<img src"mist.png" alt="Mist"/>`
+    }
+}
+showConditionIcon()*/
```
ericaglimsholt: You are very close! First of all you need this part higher up, but I guess you added it here only for now. So if you log `currentWeather` you get an error that says that `weather is not defined`. The way to go around here is to take a little piece at a time. So starting logging just `json` you can see that you need to go in to something called `list`. From then you have an array, so we need a loop in some sort, eg a foreach. Then you can build up your new variable. If you now log `currentWeather` you get all of the states for the weather and can work from there :)

```
const showConditionIcon = () => {
   json.list.forEach(element => {
      const currentWeather = element.weather[0].description
      if (currentWeather === "clear sky") {
         currentIcon.innerHTML = `<img src"sun.png" alt="Sunny"/>`
      } 
   })
}
showConditionIcon()
``` 
image

Angelicabrodin: Thank you Erica :) But where do I actually need to put this function, inside of the fetch or outside?

PR #18, opened 2019-11-03
PR - Tiago Ferreira - Weather app

tavferreira:
https://tiago-weather-app.netlify.com


script.js
```diff
@@ -0,0 +1,328 @@
...
+  "Nov",
+  "Dec"
+];
+
+//Geolocation functions
+function locationRetrieved(pos) {
+  let crd = pos.coords;
+  city = `lat=${crd.latitude}&lon=${crd.longitude}`;
+  return city;
+}
```
ericaglimsholt: You can rewrite you functions to be arrow functions (es6)! 
```
const locationRetrieved = (pos) => {
   let crd = pos.coords;
   city = `lat=${crd.latitude}&lon=${crd.longitude}`;
   return city;
}
```

script.js
```diff
@@ -0,0 +1,328 @@
...
+    case "section3":
+      document.getElementById("arrow3").classList.toggle("rotate");
+      break;
+    case "section4":
+      document.getElementById("arrow4").classList.toggle("rotate");
+      break;
+    case "section5":
+      document.getElementById("arrow5").classList.toggle("rotate");
+      break;
+  }
```
ericaglimsholt: Nice idea about this! But I think you can make this a lot more simple than what it is. If you use classes and add one class to the one that's active you can make this to 3 rows or so :)

PR #23, opened 2019-11-03
Project 6 - Weather App

Nasimmhn:
https://weather-by-nasim.netlify.com/

Here is my weather app project. It was challenging to get the days correct. However, I now understand the concept of an 'api'.


index.js
```diff
@@ -0,0 +1,80 @@
+const city = document.getElementById("city");
+const weatherIcon = document.getElementById("weather_icon");
+const temp = document.getElementById("temp");
```
ericaglimsholt: Maybe you can give this one a better variable name? Temp can be short of a lot of thing, usually used as a short of "temporary" but I guess that you mean "temperature" here

index.js
```diff
@@ -0,0 +1,80 @@
...
+    const newDate = new Date(json.list[0].dt_txt)
+    const todayDate = newDate.toLocaleDateString("en-US", options)
+
+    document.getElementById("date").innerHTML = `${todayDate}`
+
+    weather_icon.src = `https://openweathermap.org/img/wn/${json.list[0].weather[0].icon}@2x.png`
+    temp.innerHTML = `${getNumber(Math.round(json.list[0].main.temp))}&deg;C`;
+
+    sunrise.innerHTML = `${("0" + sunRise.getHours()).slice(-2)}:${("0" + sunRise.getMinutes()).slice(-2)}`;
+    sunset.innerHTML = `${("0" + sunSet.getHours()).slice(-2)}:${("0" + sunSet.getMinutes()).slice(-2)}`;
```
ericaglimsholt: Have you tried out `toLocalTime`? That would help you not needed to do a equation https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString

PR #25, opened 2019-11-04
Project weather app - Linda Isell (week 6)

liissee:
Here's my project:
https://weather-app-linda.netlify.com/ 
[https://weather-app-linda.netlify.com/](https://weather-app-linda.netlify.com/)
Really fun week but also challenging with the forecast.. 
I don't feel that I understand how to loop the api to show the 3 hours forecast which seemed like the "easy" way to do it? 

Struggles: 
- Understand the forEach to show api's
- Couldn't fix the "else" statement if the user don't accept to use geolocation? 
- How do I get the icons to show day-icons? Now they only show night version..


liissee:
Added the link. Don't know why it didnt show first time :)

script1.js
```diff
@@ -0,0 +1,159 @@
...
+                    <p>min: ${geoTempMin.toFixed(1)} C° |  max: ${geoTempMax.toFixed(1)} C°</p>
+                    
+                <p>Sunrise: ${geoSunrise}</p>
+                <p>Sunset: ${geoSunset}</p>
+                <p id="border-bottom-geo"></p>`
+            });
+    });
+} else {
+    //does not work?..
+    //h1.innerHTML = "Geolocation is not supported by this browser.";
```
JennieDalgren: I think this will never be true, the if statement `navigator.geolocation` will always be true so I'm not sure else will ever be triggered.

script1.js
```diff
@@ -0,0 +1,159 @@
...
+                return response.json();
+            })
+            .then(json => {
+                console.log(json);
+                const geoName = json.name;
+                const geoTemp = json.main.temp;
+                const geoTempMin = json.main.temp_min;
+                const geoTempMax = json.main.temp_max;
+                const geoDescription = json.weather[0].description;
+                const geoSunrise = new Date(json.sys.sunrise * 1000).toLocaleTimeString(
```
JennieDalgren: Nice work

script1.js
```diff
@@ -0,0 +1,159 @@
...
+let long;
+let lat;
+const theWeatherGeo = document.getElementById("weather-geolocation");
+
+
+if (navigator.geolocation) {
+    navigator.geolocation.getCurrentPosition(position => {
+        lon = position.coords.longitude;
+        lat = position.coords.latitude;
+
```
JennieDalgren: Cool that you've tried the geolocation stretch goal! Well done 🏆

script1.js
```diff
@@ -0,0 +1,159 @@
...
+        //difference is this one collects what ever we return (this case - "value.main.temp"). 
+        //Map goes over each objects in this array and sucks out the temperature value and creates a new array of temp.values.
+        const temps = weatherValues.map((value) => value.main.temp)
+
+        const minTemp = Math.min(...temps)
+        const maxTemp = Math.max(...temps)
+
+        //get the weather descriptipn for the forecast (not using in HTML at the moment)
+        const descriptions = weatherValues[0].weather[0].description
+        const icons = weatherValues[0].weather[0].icon
```
JennieDalgren: The reason why you're getting the dark icons is because your choosing to use the first weather value of the day, 00:00. And then you'll have that night icon.
`console.log(weatherValues)`
Check this console log and maybe it gets clearer?

script1.js
```diff
@@ -0,0 +1,159 @@
...
+//GEOLOCATION SECTION
+let long;
+let lat;
+const theWeatherGeo = document.getElementById("weather-geolocation");
+
+
+if (navigator.geolocation) {
+    navigator.geolocation.getCurrentPosition(position => {
+        lon = position.coords.longitude;
+        lat = position.coords.latitude;
```
JennieDalgren: nice that you've done the geolocation stretch goal.

script1.js
JennieDalgren: I think the if statement will always be true so the else will never be reached?
This is just checking if the browser supports geolocation.
What you want to check is if you get a position from `navigator.geolocation.getCurrentPosition()` and do different things based on that or not.

script1.js
liissee: Thanks! When I change to `const icons = weatherValues[4].weather[0].icon` I get the daily icons but I also get an error message in .catch saying:

Skärmavbild 2019-11-18 kl  12 49 06

why is that..?

PR #26, opened 2019-11-04
https://trusting-golick-5d6863.netlify.com/

horsefacegirl:


script.js
```diff
@@ -0,0 +1,95 @@
...
+.then(json => {
+  console.log(json)
+  json.list.forEach(day => {
+    const time = new Date(day.dt * 1000)
+
+    if (time.getUTCHours() === 12) {
+      const weekday = new Intl.DateTimeFormat("en-US", { weekday: "long" }).format(time)
+
+      theWeekdays.innerHTML += `<p>${weekday}</p>`
+      weekdayTemp.innerHTML += `<p>${Math.floor(
```
JennieDalgren: Careful here, `Math.floor()` will always round the number down to nearest integer.
i.e `Math.floor(3.2) => 3` and `Math.floor(3.9) => 3` 
Maybe you should use `Math.Round()` instead.
[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round)

PR #31, opened 2019-11-04
Project: Weather app, Michel Elmoznino Laufer

MichelLaufer:
Hi! 
I have attached the Netlify-link to the weather app here: 
https://michel-weather.netlify.com/ 

Best,
Michel


script.js
```diff
@@ -0,0 +1,138 @@
...
+}
+
+
+const stockholmWeather = () => {
+  const forecastDiv = document.getElementById('forecast')
+  document.getElementById("weatherTitel").innerHTML=''
+  document.getElementById("stockholmWeather").innerHTML=''
+  document.getElementById("romeWeather").innerHTML=''
+  document.getElementById("barcelonaWeather").innerHTML=''
+  document.getElementById("parisWeather").innerHTML=''
```
ericaglimsholt: Just because you are doing this multiple times it would be better if you add this to a function and call the function instead. That would mean less code = always good

PR #53, opened 2020-02-19
Weather App Anna Takman

annatakman:
Link to my deployed project: https://annas-weather-app.netlify.com 

My reflections on how this week's project turned out: This was the hardest project for me so far. It was challenging to understand the API and what data to get from it, and how to get it and present it. 

Things I'd like to have clarified or explained in more detail: I would like to know a nicer CSS-solution for how to get the columns with weekdays, temperatures and weather icons in the five day forecast aligned. I just used margin to get them aligned but I guess you should be able to use flexbox to solve it.


index.html
```diff
@@ -0,0 +1,41 @@
...
+      <h1 id="temperature"></h1>
+      <div id="weatherPic" class="weatherPic"></div>
+    </section>
+
+    <section>
+      <h2 id="city"></h2>
+      <h3 id="description"></h3>
+    </section>
+
+    <section id="sunriseAndSunset" class='sunriseAndSunset'>
```
ericaglimsholt: You are using both `"` and `'` in this file. It would be good if you choose one of them and go for that instead of mixing them :)

script.js
```diff
@@ -0,0 +1,83 @@
...
+  .then((json) => {
+    temperature.innerHTML = `${json.main.temp.toFixed(0)}<sup>&#8451</sup>`
+    city.innerHTML = json.name
+    description.innerHTML = json.weather[0].description
+    const sunriseTime = new Date(json.sys.sunrise * 1000).toLocaleTimeString([], { hour: '2-digit', minute: '2-digit' })
+    sunrise.innerHTML = `sunrise ${sunriseTime}`
+    const sunsetTime = new Date(json.sys.sunset * 1000).toLocaleTimeString([], { hour: '2-digit', minute: '2-digit' })
+    sunset.innerHTML = `sunset ${sunsetTime}`
+
+    if (json.weather[0].id === 804) {
```
ericaglimsholt: Nice that you have a "theming" going on with different images. If you want to develop this more you could create a function for getting the images so you don't need all of the if-statements :D

PR #58, opened 2020-02-20
Weather app Camilla 

camillagaus:
Link to my deployed project: https://nifty-albattani-3edf5f.netlify.com/

My reflections on how this week's project turned out: It was a bit at the beginning, I think it was a bit confusing how to access the data from the weather site but after som trial and error I got hold of the concept and how to use it. It was fun to create something "real" with data that changes when the weather change. 

Things I'd like to have clarified or explained in more detail: I would like to know more about how to place the ${city} in the API adress and how to work with, instead of inserting different API addresses for every city. I would also want to learn more about how to talk about what we are doing with the correct terms


code/weather.js
```diff
@@ -0,0 +1,100 @@
+const container = document.getElementById('city')
```
ericaglimsholt: Maybe `containerCity` to follow the other name structure?

code/weather.js
```diff
@@ -0,0 +1,100 @@
...
+const containerForecast = document.getElementById('forecast')
+const currentCondition = document.getElementById('weather')
+const conditionImg = document.getElementById('weatherImage')
+const countries = document.getElementById('otherCountries')
+
+const uppsala = 'https://api.openweathermap.org/data/2.5/weather?q=Uppsala,Sweden&units=metric&APPID=ed2d4804e9f1b16e037cdf4e25100bc4'
+const uppsalaForecast = 'https://api.openweathermap.org/data/2.5/forecast?q=Stockholm,Sweden&units=metric&APPID=ed2d4804e9f1b16e037cdf4e25100bc4'
+const alicante = 'https://api.openweathermap.org/data/2.5/weather?q=Alicante,Spain&units=metric&APPID=ed2d4804e9f1b16e037cdf4e25100bc4'
+const london = 'https://api.openweathermap.org/data/2.5/weather?q=London,UK&units=metric&APPID=ed2d4804e9f1b16e037cdf4e25100bc4'
+const tampa = 'https://api.openweathermap.org/data/2.5/weather?q=Tampa,US&units=metric&APPID=ed2d4804e9f1b16e037cdf4e25100bc4'
```
ericaglimsholt: Here is where you can create a function, as you wanted to have described. Let's say that you create a general function for getting the weather where you add a city as a prop:
```
const getWeather = city => {
  return `https://api.openweathermap.org/data/2.5/weather?q=${city},Sweden&units=metric&APPID=ed2d4804e9f1b16e037cdf4e25100bc4`
}

getWeather('Uppsala')
```

code/weather.js
```diff
@@ -0,0 +1,100 @@
...
+
+fetch(uppsala)
+.then((response) => {
+  return response.json()
+})
+
+.then((json) => {
+  container.innerHTML = `<h1>${json.name}</h1>`
+  let tempAsJSON = json.main.temp.toFixed(0)
+  containerCelcius.innerHTML += `<h1> ${tempAsJSON}${'&#730;'}<sup>C</sup>  </h1>`
```
ericaglimsholt: Nice that you are using `<sup>` element!

code/weather.js
```diff
@@ -0,0 +1,100 @@
...
+  let windSpeed = json.wind.speed
+  const weatherSymbol= json.weather[0].icon
+  conditionImg.innerHTML += `<img class="weatherSymbol" src="${weatherSymbol}.png" alt="">`
+  let sunrise = new Date(json.sys.sunrise * 1000).toLocaleTimeString([], {hour: '2-digit', minute: '2-digit'})
+  let sunset = new Date(json.sys.sunset * 1000).toLocaleTimeString([], {hour: '2-digit', minute: '2-digit'})
+ containerSunsetRise.innerHTML += `<h4> Sunrise: ${sunrise} <img class="up-down" src="sunrise.png"> </h4> <h4>Sunset: ${sunset} <img class="up-down" src="sunset.png"> </h4>`
+ 
+
+ var now = new Date();
+ var short = now.toLocaleTimeString([], {timeStyle:'short'})
```
ericaglimsholt: Try to use `const` or `let` before `var` if possible!

PR #62, opened 2020-02-20
project-weather-app Emma

emmapre:
Link to my deployed project:
https://emmas-weather-app.netlify.com/

I'm quite proud of the app, and I'm amazed of how much it is possible to do with APIs. It was fun! There are some styling things that I would have liked to fixed and I think I know how it could be done, but I felt I did not have the time for it right now. I should also have put the styling in css and used classes instead of styling in the js, but I realized that too late.  I think I could also have put more things into variables instead, to clean up the code. 

I would like to know where it would have been better to make a variable of something and where I could have done a function for something, I feel that I am not yet ablet to see where I can simplify the code.


code/script.js
```diff
@@ -0,0 +1,126 @@
...
+    sunriseContainer.innerHTML = `<h3>sunrise ${sunriseTime}</h3>`
+
+    const sunset = new Date(json.sys.sunset * 1000)
+    const sunsetTime = sunset.toLocaleTimeString([], {
+      timeStyle: 'short'
+    });
+
+    sunsetContainer.innerHTML = `<h3>sunset ${sunsetTime}</h3>`
+
+    if (json.weather[0].main === "Rain" || json.weather[0].main === "Drizzle") {
```
ericaglimsholt: If you add the `<h1>` text in a variable you can create a function for this like
```
const getWeatherTheme = weatherType => {
  placeContainer.innerHTML = weatherType.heading
  weatherIcon.src = weatherType.icon
  bigContainer.style.background = weatherType.backgroundColor
  bigContainer.style.color = rainy.weatherType
}

getWeatherTheme('rain')
```

code/style.css
```diff
@@ -0,0 +1,91 @@
...
+
+h3 {
+  font-weight: 400;
+  margin-top: 10px;
+  margin-bottom: 10px;
+}
+
+#weather-icon {
+  width: 100px;
+  animation-name: floating;
```
ericaglimsholt: Love this!!

PR #65, opened 2020-02-20
project-weather-app complete - Elsa Reuterswärd

elsareutersward:
Link to my deployed project: 
https://compassionate-hoover-547cdb.netlify.com

I had so much trouble with the project for the first few days, and couldn't work out how to write my fetches probably. After a lto of googling and help, I manage to work it out and then I got really in to the project and using everything we learnt in the last couple of weeks by using objects, arrays, animations, toggles etc... So in the end it was a very fun and educational project!

I still have some problem of understanding how to think/structure it when I write javascript. I understand when it's explained to me, but I still find it really difficult to write all on my own.  I rarely get any of my javascript code to work without getting help...


index.html
```diff
@@ -0,0 +1,178 @@
...
+        <div id="background-cloud">
+            <div class="x1">
+                <div class="cloud"></div>
+            </div>
+            <div class="x2">
+                <div class="cloud"></div>
+            </div>
+        </div>
+
+        <div id="background-rain">
```
ericaglimsholt: I loooooove this!!!

PR #72, opened 2020-02-21
Project 6 - Johanna Åstrand

jastrand:
Link to my deployed project:
https://yo-weather.netlify.com/

The first part of the project was really fun and easy to understand. The weather forecast was complicated to understand and code since I had to use many things I have ever done before. I found a lot of information and code suggestions online, but I think I would like to come back to this part in the future to write the code smarter and better. I would also like to spend more time on the design to change color depending on weather as well as adding different cities.

I think the most complicated for me is to find a smart way to divide the project into different parts and knowing what techniques I should use and how to structure it


script.js
```diff
@@ -0,0 +1,79 @@
...
+
+// TODAYS WEATHER
+
+fetch('https://api.openweathermap.org/data/2.5/weather?q=Stockholm,Sweden&units=metric&APPID=10c15495461885dfddf7c2f3846d4e30')
+  .then((response) => {
+    return response.json()
+  })
+
+  .then((json) => {
+    console.log(json)
```
ericaglimsholt:
```suggestion
```

PR #73, opened 2020-02-21
Project Weather App Lisa sturegård

LisaStur:
Link to my deployed project: 
https://gifted-golick-886bea.netlify.com/

Lots to learn in this project, super! Wish I would have found a good solution for when the geolocator is not working, or is denied by the user but ran out of time...


code/index.html
```diff
@@ -0,0 +1,29 @@
...
+  <header>
+    <section id="headerPicture" class="weather-image"></section>
+    <section id="headerTemp" class="header-temp"></section>
+  </header>
+  <section class="main-container">
+    <section id="today" class="today">
+      <section id="weather" class="weather">
+        Loading...
+      </section>
+      <section id="SunRiseSet" class="sun-rise-set"></section>
```
ericaglimsholt: Your naming convention in other places begins with a small letter first, maybe this should be the same?
```suggestion
      <section id="sunRiseSet" class="sun-rise-set"></section>
```

code/weather.js
```diff
@@ -0,0 +1,79 @@
...
+const locationDropdown = document.getElementById('locationDropdown')
+const  weatherIcon = document.getElementById('weatherImage')
+
+
+if (navigator.geolocation) {
+  navigator.geolocation.getCurrentPosition(displayLocationInfo);
+}
+function displayLocationInfo(position) {
+  const lng = position.coords.longitude;
+  const lat = position.coords.latitude;
```
ericaglimsholt: I don't know about this naming. In other places it can be hard to understand what `lng` means

LisaStur:
Big thanks Erica! Yes, should be only one css file, thanks for spotting! And the other comments, very good feedback will update!

PR #76, opened 2020-02-21
Johannas weather app

olsenjohanna:
Link to my deployed project: https://suspicious-jepsen-1af114.netlify.com/

My reflections on how this week's project turned out: This week have been so fun! I'm pretty happy with how it turned out even though there where some things I did not manage to do (for example min/max temp, also some styling that I did not really know how to get to work in comibnation with javascript and changing cities. But overall I'm pretty happy, it feels that I start to understand some more about javascript! 

Things I'd like to have clarified or explained in more detail: When looking on my weather app on the phone I get "invalid date". How can I do to make that work?


code/script.js
```diff
@@ -0,0 +1,173 @@
...
+// if statement to select background depending on day/night
+if (hour >= 7 && hour <= 19) {
+  body.classList.remove('night')
+  body.classList.add('day')
+} else if (hour > 19 && hour <= 24) {
+  body.classList.add('night')
+  body.classList.remove('day')
+} else if (hour >= 00 && hour < 7) {
+  body.classList.add('night')
+  body.classList.remove('day')
```
ericaglimsholt: These two seems to do the same thing so your can use the same else-if here! Use `||`(= or) between the different statements
```
} else if (hour > 19 && hour <= 24 || hour >= 00 && hour < 7) {
  body.classList.add('night')
  body.classList.remove('day')
}
```

code/script.js
```diff
@@ -0,0 +1,173 @@
...
+        const windRounded = Math.round(wind * 10) / 10
+
+        // Set variable for icon
+        const iconIdForecast = item.weather[0].icon
+        const iconForecast = `http://openweathermap.org/img/wn/${iconIdForecast}.png`
+
+        // Print information to DOM
+        forecastHeader.innerHTML = `<h3> 5 day forecast</h3>`
+        forecastDay.innerHTML += `<p>${day}</p>`
+        forecastIcon.innerHTML += `<img src=${iconForecast} />`
```
ericaglimsholt: Missing the alt here on the image!

PR #127, opened 2020-09-27
Weather App welcomeweather

bloeli:
Link to my deployed project:
https://welcomeweather.netlify.app/

Think it wen't well even though I wan't to learn more, have more time to learn and do more :)

I noticed (after pull requested) after 22.00 same day that the night icon did not appear (see file: weather.js --> 59). Earlier on the day I tested it writing (currentTime >= 14 && currentTime <= 16) and the night icon appeared. I have tested this, but when I write a number which starts with 0, like 04 or 06, or without 0, like 4 or 5 it does not work. If I write 16, 17, 22, 23 etc it works. Please help me!


code/weather.js
```diff
@@ -0,0 +1,85 @@
...
+    const sunsetTime = sunset.toLocaleTimeString([], { timeStyle: "short" });
+    sunriseInfo.innerHTML += `<h2> Sunrise ${sunriseTime}</h2>`;
+    sunsetInfo.innerHTML += `<h2> Sunset ${sunsetTime} </h2>`;
+});
+
+//Function that show the night icon when it is nighttime (22-04), and remove the weather icon & text so it doesn't show & adding a text to show up with the night icon:
+const night = () => {
+    const currentTime = new Date().getHours();
+    console.log(currentTime)
+    if (currentTime >= 22 && currentTime <= 04){
```
bloeli, line 59: I noticed (after pull requested) after 22.00 same day that the night icon did not appear. Earlier on the day I tested it writing (currentTime >= 14 && currentTime <= 16) and the night icon appeared. I have tested this, but when I write a number which starts with 0, like 04 or 06, or without 0, like 4 or 5 it does not work. If I write 16, 17, 22, 23 etc it works. Please help me!

code/index.html
```diff
@@ -0,0 +1,43 @@
...
+                    <img class="weather-icon" id="weatherIcon"  src="">
+                </div>
+                <img class="night-icon" id="nightIcon" src="">
+                <div id="weatherDescription"></div>
+            </div>
+            <div class="header-part-2">
+                <div id="city"></div>
+                <div  id="mainTemp"></div>
+            </div>
+        </header>
```
Corneliaryas, line 24: An interesting approach to have this first part of the current weather in a header tag and the rest of the current weather information (weather text and sunrise/sunset) in the main tag, could it be a bit easier to follow the code if both these parts were in the same bigger part (a section-tag maybe)?

code/index.html
```diff
@@ -0,0 +1,43 @@
...
+        <main>
+            <div class="weather-text" id="weatherText"></div>
+            <div class="sunrise-sunset">
+                <div id="sunriseInfo"></div>
+                <div id="sunsetInfo"></div>
+            </div>
+            <!--FORECAST-->
+                <div class="forecast">
+                    <div id="theWeekdays"></div>
+                    <div id="forecastMainTemp"></div>
```
Corneliaryas, line 34: Great for performance to have the html tags already in the HTML file instead of creating all of it in JS I think!  👍

code/style.css
```diff
@@ -0,0 +1,185 @@
...
+@keyframes example {
+  0%   {background-color: rgb(12, 192, 247);}
+  25%  {background-color: rgb(247, 12, 208);}
+  50%  {background-color: rgb(187, 8, 187);}
+  75%  {background-color: rgb(247, 173, 12);}
+  100% {background-color: rgb(255, 115, 0);}
+}
+
+.header-part-1 {
+  max-width: 400px;
```
Corneliaryas, line 18: The max-width is set to the same pixel value multiple times throughout the css, could it be an idea to set the parent(s) to that max-width (for instance .current weather and main, or even the body tag itself) and then just have width: 100% on the other ones, then you would only have to change the pixel value once in the media queries? :)

code/style.css
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  footer {
+    max-width: 4000px; 
+    margin: 140px;
+    font-size: 32px;
+    text-align: center;
+    border-top: white solid;
+  }
+}
```
Corneliaryas, line 185: Easy to follow CSS with great order of declarations!

code/weather.js
```diff
@@ -0,0 +1,85 @@
...
+            document.getElementById('weatherIcon').src = './icons/iconfinder_Snow_3741358.svg';
+            weatherText.innerHTML += "<h2>Stay warm and cosy inside? Out skiing or build a snow man?</h2>"
+        } else {
+            document.getElementById('weatherIcon').src = './icons/iconfinder_Weather_forecast_weather_station_radar_6646328.svg';
+            weatherText.innerHTML += "<h2>Welcome to this weather app!</h2>"
+        }
+    }
+    //Calling the function:
+    weatherIconShow(json.weather[0].main)
+    //Setting sunrise and sunset to show on website with time hours and minutes:
```
Corneliaryas, line 46: Super nice with the comments, makes it easy to follow along!

code/weather.js
```diff
@@ -0,0 +1,85 @@
+//API:s WITH WEATHER INFO TO FETCH:
+const apiUrl = 'https://api.openweathermap.org/data/2.5/weather?q=Stockholm,Sweden&units=metric&APPID=8b97619989976c72fc1e602d8c793890';
+const apiUrl2 = 'https://api.openweathermap.org/data/2.5/forecast?q=Stockholm,Sweden&units=metric&APPID=8b97619989976c72fc1e602d8c793890';
```
Corneliaryas, line 3: Consider the naming of variables here, if you had more URLs it might be hard to follow which one is doing what. Maybe something like apiUrlForecast instead of apiUrl2?

Corneliaryas (approved):
Great job Elisa! It shows that you have put time on this, you have clean code and a great site. I really like the icons and the messages that change depending on the weather type! 😃👍

code/weather.js
bloeli: Thank You it worked :) First I had || but it did'nt work, it must have been something other then that was'nt correct, it looked like this:
 const night = () => {
    const currentTime = new Date().getHours();
    if (currentTime >= 16 || currentTime <= 17) {
        document.getElementById('nightIcon').src = './icons/iconfinder_03_moon_sleepy_night_emoticon_weather_smiley_3375686.svg';
    } else {
        //INGET';
    } 

Maybe it was somethinh with the else statement that did not make the function work before .

bloeli: * and I also tested/changed it to 22 and 04 instead of 16 and 17 of course

code/README.md
```diff
@@ -0,0 +1,52 @@
...
+- SVG icons/pictures
+- Google searching for information, free icons, etc
+- Teammates helping eachother <3
+
+## PLAN:
+1. Read through every week material from Technigo & see the movies
+2. Fork the project in GitHub
+3. Team Lab Session that was meaningful/learnfull for this individual project
+4. Concentrating step by step on blue level, then adding more if time. When step 5 (style the weather app) I did a sketch on how I would like it to look, painted also the element/boxes so that I could easily think on how to style and do the html and css.
+
```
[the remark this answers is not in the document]
bloeli: Thank You!

code/style.css
```diff
@@ -0,0 +1,185 @@
...
+    animation-duration: 16s;
+}
+
+@keyframes example {
+  0%   {background-color: rgb(12, 192, 247);}
+  25%  {background-color: rgb(247, 12, 208);}
+  50%  {background-color: rgb(187, 8, 187);}
+  75%  {background-color: rgb(247, 173, 12);}
+  100% {background-color: rgb(255, 115, 0);}
+}
```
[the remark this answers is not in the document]
bloeli: Thanks :)

code/index.html
```diff
@@ -0,0 +1,43 @@
...
+                </div>
```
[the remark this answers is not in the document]
bloeli: Absolutely :) fixed!

code/style.css
bloeli: Good idea! I changed to that :)

code/weather.js
bloeli: Good point! I change this as well :)

code/weather.js
```diff
@@ -0,0 +1,84 @@
...
+    sunsetInfo.innerHTML += `<h2> Sunset ${sunsetTime} </h2>`;
+});
+
+//Function that show the night icon when it is nighttime (22-04), and remove the weather icon & text so it doesn't show & adding a text to show up with the night icon:
+
+const night = () => {
+    const currentTime = new Date().getHours();
+    if (currentTime >= 22 || currentTime <= 04) {
+        document.getElementById('nightIcon').src = './icons/iconfinder_03_moon_sleepy_night_emoticon_weather_smiley_3375686.svg';
+        weatherText.innerHTML +="<h2>Hello there Night Owl! Can't sleep? Cathing stars? Dancing?</h2>";
```
EmOlsz, line 61: ;)

code/weather.js
```diff
@@ -0,0 +1,84 @@
...
+        var element = document.getElementById("weatherIconParent");
+        element.parentNode.removeChild(element);
+    }
+}
+night();
+//Calling the function for showing the night icon + text
+
+//FORECAST FOR 5 DAYS:
+////Fetch the forecast weather API:
+fetch(apiForecastWeather).then((response) => {
```
EmOlsz, line 71: After first then we should break line

code/weather.js
```diff
@@ -0,0 +1,84 @@
...
+});
+
+//Function that show the night icon when it is nighttime (22-04), and remove the weather icon & text so it doesn't show & adding a text to show up with the night icon:
+
+const night = () => {
+    const currentTime = new Date().getHours();
+    if (currentTime >= 22 || currentTime <= 04) {
+        document.getElementById('nightIcon').src = './icons/iconfinder_03_moon_sleepy_night_emoticon_weather_smiley_3375686.svg';
+        weatherText.innerHTML +="<h2>Hello there Night Owl! Can't sleep? Cathing stars? Dancing?</h2>";
+        var element = document.getElementById("weatherIconParent");
```
EmOlsz, line 62: Why var?

code/weather.js
```diff
@@ -0,0 +1,84 @@
...
+const apiCurrentWeather = 'https://api.openweathermap.org/data/2.5/weather?q=Stockholm,Sweden&units=metric&APPID=8b97619989976c72fc1e602d8c793890';
+const apiForecastWeather = 'https://api.openweathermap.org/data/2.5/forecast?q=Stockholm,Sweden&units=metric&APPID=8b97619989976c72fc1e602d8c793890';
+
+//GLOBAL SCOPE INFO TO WEATHER:
+const weatherDescription = document.getElementById('weatherDescription');
+const weatherText = document.getElementById('weatherText');
+
+//CURRENT WEATHER:
+//Fetch the current weather API:
+fetch(apiCurrentWeather).then((response) => {
```
EmOlsz, line 11: After first then we should break line

EmOlsz (approved):
Well done project Elisa :) I'm big fan of your design, functionality is also written in a good way. Congratulations!
Take a look on my comments, they might be helpful for future projects :)

code/style.css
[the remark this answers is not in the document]
bloeli: Thank you :)

code/style.css
```diff
@@ -0,0 +1,180 @@
+body {
+    font-family: 'Montserrat';
+    color: white;
+    background-color: rgb(12, 192, 247);
+    animation-name: example;
+    animation-duration: 16s;
```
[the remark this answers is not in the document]
bloeli: Thank you :) I'll change to that :)

code/weather.js
bloeli: What do you mean? :)

bloeli: > After first then we should break line
What do you mean? :)

code/weather.js
bloeli: Why not? No, kidding, that's a good question.... I found this solution google:ing and then I did not change it, but 'I'll change it to const now.

code/weather.js
EmOlsz: It should rather look like this :

```
fetch(apiForecastWeather)
    .then((response) => { ... })
    .then((data)) => { ... });
```

but you wrote it like this :

```
fetch(apiForecastWeather).then((response) => { ... })
    .then((data)) => { ... });
```

code/weather.js
EmOlsz: We shouldn't use `var` if there is no exact reason for that : 

1. `const` should be always your default option
2. If you see that you need to overwrite your variable later on, then use `let`
3. If you see that you need to create your variable in some kind of if statement AND (very important) you will need to access this variable outside of if statement, then use `var`

In your case you are declaring variable in if statement, that's true, but you never use it outside of it, so it's completely ok to use `const` :)

PR #133, opened 2020-09-27
Abstract Weather app by Bam

BamStugendal:
**Link to my deployed project:**
https://abstract-weather-app.netlify.app/code/

**My reflections on how this week's project turned out:**
I like how the design looks and it was fun to do a project you could use in everyday life with real information. I wanted to do the black level and started coding the geolocation function. But didn't have time to finish it because I had less time this week to work with this project. 

**Things I'd like to have clarified or explained in more detail:**
Maybe geolocation and how to use it?


Code/index.html
```diff
@@ -0,0 +1,81 @@
...
+            </div>
+            <div class="day-container">
+                <p class="day">day</p>
+                <p class="temperature">N/A °C</p>
+            </div>
+            <div class="day-container">
+                <p class="day">day</p>
+                <p class="temperature">N/A °C</p>
+            </div>
+    </section>
```
Thessan, line 68: Another approach could be to only create one div and paragraph for all five days, and then add variables in the forEach function in JS.

I think this might be a good approach if you later would like create for example a ten day forecast, because then you don't have to keep track of how many div's you have in HTML, there would be no need to create new ones (code coaches, please correct me if I'm mistaken in this thought) =)

Example:

```
<div class="forecast">
        <p id="five-days"></p>
    </div>
```

```
filteredForecast.forEach((forecast) => {
            const temp = forecast.main.temp;
            const date = new Date(forecast.dt_txt);
            const options = { weekday: 'short' };
            const localDateString = date.toLocaleDateString('en-EN', options);
            const fiveDays = document.getElementById('five-days');
   
            fiveDays.innerHTML += `<p>${localDateString}   ${temp.toFixed(1)}°c</p>`; 
```

Code/script.js
```diff
@@ -0,0 +1,163 @@
...
+
+//Change background depending on time
+const changeBackground = (currentWeather) => {
+    const currenttime = currentTime();
+    const sunrise = generateTime(currentWeather.sys.sunrise)
+    const sunset = generateTime(currentWeather.sys.sunset)
+
+    document.body.style.backgroundImage = currenttime > sunrise && currenttime < sunset 
+    ? 'linear-gradient(190deg, #FFC6C8 0%, #FAD56A 100%)'
+    : 'linear-gradient(190deg, #ffc6c8 0%, #976bb6 100%)'
```
Thessan, line 134: I think these are very nice colors, I like the linear-gradient :star:

Code/style.css
```diff
@@ -0,0 +1,249 @@
...
+    flex-direction: row;
+    justify-content: space-between;
+    width: 90%;
+    box-sizing: border-box;
+    text-transform: capitalize;
+    border-bottom: 2px #fff dashed;
+    margin-bottom: 0px; 
+    font-weight: 400;
+    margin: 0 auto;
+}
```
Thessan, line 205: Maybe think about being a bit more consistent with your code structure. I mean, try to keep font details, margin, padding etc in the same order for every element :blush:

Code/style.css
```diff
@@ -0,0 +1,249 @@
...
+
+@media only screen and (min-width: 1024px) {
+    .weather-type {
+        font-size: 80px;
+    }
+
+    .todays-temp {
+        left: 30%;
+    }
```
Thessan, line 248: Nice that you don't have to change that much code to make it look good :clap:

Thessan (commented):
I really like your weather app. The colors and layout is very appealing. I first thought your cloud and rain drops were images, so it was cool to see they were made of divs. Well done! :clap:

EmOlsz (approved):
Very nice project, I'm impressed A LOT! Keep it this way and don't stop :) It's vert well structured and what's most important - consequent. That's one of most important aspects, you will understand that when bigger, more complex projects will enter.
Please, take a look on my comments, they are mostly minor, but still important.
Good job :)

PR #158, opened 2021-02-19
Technigo Boot Camp Week 6: Weather App (pair-programming project)

idasylven:
This was a project pair-programmed by Anna Lenksjö and Ida Sylvén.

Link to our deployed project here:

https://cranky-euler-1b243b.netlify.app/

Our reflections on how this week's project turned out:

We are very happy with our final project! We fulfilled all blue level requirements and particular attention to replicating one of the two design option given to us. All the design features we have changed compared to the original sketch have been changed with intent. If we have more time we would have liked to focus even further on the various layout displays for bigger screens. 

We tried our best to create the exact shape of the lower edge of the main section, the original is even more rounded than the shape we ended up with. We aslo had issues creating rounded edges on our "Detailed forecast"-button using border radius, we can tell this element has inherited a padding that we cant seem to get rid off. As we progressed we realized it becomes harder and harder to name classes and id:s!


style.css
```diff
@@ -0,0 +1,270 @@
...
+        display: none;
+    }
+    /***************** Hamburger menu *****************/
+    /* hiding the checkbox*/
+    
+    #menu {
+        visibility: hidden;
+        display: flex;
+        align-items: flex-start;
+    }
```
ElleFrankenberg, line 34: We believe that you should style be class and not by id.

style.css
```diff
@@ -0,0 +1,270 @@
...
+    
+    label,
+    label::before,
+    label::after {
+        position: absolute;
+        width: 28px;
+        height: 3px;
+        border-radius: 20px;
+        background: white;
+    }
```
ElleFrankenberg, line 45: really nice css structuring that you styled several elements at the same time.

ElleFrankenberg (approved):
Over all the whole page looks very professional! Really nice structured code and indentation. The button to see more detaild forecast is not working, we are not sure that it was supposed to? Otherwise really really good job with the project✨ ✨ ✨ ✨ ✨ ✨

style.css
KarolCislowski: Yes, styling with ids can make a lot of problems in a bigger project.

style.css
```diff
@@ -0,0 +1,270 @@
+    body {
```
KarolCislowski, line 1: to reset all default margins/paddings you can use `* {margin: 0; padding: 0;}` but be careful with it because you need to provide this parameter to all your styles after doing it.

style.css
```diff
@@ -0,0 +1,270 @@
...
+        margin: auto;
+        background-color: #ffffff;
+        font-family: Roboto;
+    }
+    /***************** Main section *****************/
+
```
KarolCislowski, line 7: Your "today's" part of the app is aligned to left, and the forecast is centred. It makes it look off on bigger screens.

KarolCislowski (approved):
This is a very nice and clean project 🌈 
The styling is very nice 🎨  Please check the alignment of your sections.

Great Job!

PR #161, opened 2021-02-19
Weather App - Erika Porath, Johan Löfgren, Åsa Hildebrand

AsaHildebrand:
Link to my deployed project:
https://elastic-mirzakhani-e20ab1.netlify.app/

This weeks project went well, our teamwork has been very good and we made sure that everybody understood all of the code.

Nothing special

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,4 +9,4 @@ If we had more time we would have added diffrent cities to choose between and ch
 
 ## View it live
 
-Every project should be deployed somewhere. Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
+https://jolly-bohr-475fff.netlify.app/
--- a/README.md
+++ b/README.md
@@ -9,4 +9,4 @@ If we had more time we would have added diffrent cities to choose between and ch
 
 ## View it live
 
-https://jolly-bohr-475fff.netlify.app/
+https://elastic-mirzakhani-e20ab1.netlify.app/
--- a/README.md
+++ b/README.md
@@ -1,12 +1,11 @@
 # Weather App
 
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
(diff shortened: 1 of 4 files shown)


petercreations (approved):
Combined peer code review on Erika's, Johan's and Åsa's project week 6 by Shaun and Linnéa

Status should be APPROVED

**General Comments:**

- Clean and good-looking design! Nice idea for the tablet /desktop version.
- The code seems to follow the guidelines with correct indentations and good naming.
- You have reached the blue goals + added visual indicators of the weather.

**Readme**

- Weather App: Perhaps mention the goal to fetch data from an API as well
- Live Site: It is also possible to include an embedded link using this sample markdown syntax: `[Web Site Text](https://website-linkcom)`

**HTML**

- Line 11-12: The section tag should probably not be nested inside another section tag? `Section class=container `should probably just be a div ?
- According to W3C's HTML documentation: &quot;A section is a thematic grouping of content, typically with a heading.&quot;

**JS**

- It's working! Well done!
- Good that you use arrow functions and that you use const rather than let.
- Some comments about what different code blocks do would be useful if someone else was going to work on your code later.
- Line 6: Some ideas for different versions would be to store the city location ?q in a variable to interchange more easily using template literals or to store the full API link in a variable for a shorter API Url if it was to be used in multiple places.
- Line 14: Well done on doing the time zone off set! 😀
- Line 25: Great idea adding a catch for errors.
- Line 36: a `forEach` method could also be used to achieve the 5day forecast display instead of a for loop

**CSS**

- CSS looks efficient and follows the Don't Repeat Yourself-rule. 😀
- Color is defined both as hex color and RGB, maybe it looks better to stick to one of them?
- The 5 day weather forecast icons could be cantered using a Grid column layout

KarolCislowski (approved):
This is a very nice project 🌈 

Remember that you can always come back to it when you have some spare time and add more functionalities to make it an even greater portfolio asset! 🦄

PR #166, opened 2021-02-20
Project Weather App - Sofia Arzt Wallén, Elaine Bergström and Lars Söderman

lars-soderman:
Link to my deployed project:
https://the-best-weather-app.netlify.app/

Here is the pull request for this project!


script.js
```diff
@@ -0,0 +1,67 @@
...
+const sunset = document.getElementById('sunset')
+const topContainer = document.getElementById('top-container')
+const weeklyForecastContainer = document.getElementById('weekly-forecast-container')
+
+
+// API related variable 
+const currentWeatherApiUrl = 'https://api.openweathermap.org/data/2.5/weather?q=Stockholm,Sweden&units=metric&APPID='
+const forecastApiUrl = 'https://api.openweathermap.org/data/2.5/forecast?q=Stockholm,Sweden&units=metric&APPID='
+const apiKey = '3addfde144e16d817dcc3a5e9a46ea59' 
+
```
magicvilla, line 17: Just Curious 🤓 why did you save the api key as a variable?

index.html
```diff
@@ -0,0 +1,35 @@
...
+        </div>
+        <div class="sunrise-sunset">
+            <div id="sunrise"></div>
+            <div id="sunset"></div>
+        </div>
+        <div class="divider"></div>
+    </section>
+    <section class="forecast-container">
+        <div id="weekly-forecast-container"></div>   
+    </section>
```
magicvilla, line 31: Do you need all of the `div`? It seems like they are only used for DOM selectors to use them for innerHTML. Could you use maybe less `div `to do the same thing? ☺️ Maybe the `section` or `div` that act as containers and have a class that you style are enough?

index.html
```diff
@@ -0,0 +1,35 @@
...
+    <link rel="stylesheet" href="style.css">
+    <link
+    href="https://fonts.googleapis.com/css2?family=Montserrat:ital,wght@0,100;0,200;0,300;0,400;0,500;0,600;0,700;0,800;0,900;1,100;1,200;1,300;1,400;1,500;1,600;1,700;1,800;1,900&display=swap"
+    rel="stylesheet"
+    />
+    <title>Weather app</title>
+</head>
+<body>
+    <section class="top-container" id="top-container">
+        <div class="local-information">
```
magicvilla, line 16: Think about using semantic tags 🤗

style.css
```diff
@@ -0,0 +1,123 @@
...
+    font-style: bold;
+    margin: 24px 0;
+}
+
+.temperature-now {
+    font-size: 60px;
+    font-style: bold;
+}
+
+.condition p {
```
magicvilla, line 42: Try to add classes to elements that you want to style. 💌

style.css
```diff
@@ -0,0 +1,123 @@
+* {
+    margin: 0; 
+}
+
+body {
```
magicvilla, line 5: Try to add classes to elements that you want to style. 💌

style.css
```diff
@@ -0,0 +1,123 @@
...
+    flex-direction: column;
+    border: 1px solid black;
+    border-top:none;
+    border-radius: 0 0 20px 20px;
+    margin: 0 auto;
+    background: linear-gradient(to bottom right , #F47775, #F4F7F8);
+    padding: 20px 0;
+}
+
+.divider {
```
magicvilla, line 71: What does `.divider` do? Could we add a more descriptive class name? 👼

style.css
```diff
@@ -0,0 +1,123 @@
...
+    justify-content: space-evenly;
+    overflow: hidden;
+    margin: 0 40px;
+}
+
+.week-day-name {
+    width: 80px;
+}
+
+.day img {
```
magicvilla, line 91: Could you add a class name to the image instead? 😺

style.css
```diff
@@ -0,0 +1,123 @@
...
+    width: 60px;
+    height: 100%;
+    object-fit: cover;
+    margin-right: 50px;
+}
+
+/* MEDIA QUERIES */
+
+@media (min-width: 768px) {
```
magicvilla, line 100: Nice and clean media queries 😄

script.js
```diff
@@ -0,0 +1,67 @@
...
+    })
+    .catch ((err) => {
+        topContainer.innerHTML = `caught error: ${err}`
+    })
+
+// This function returns the name of day for the next 5 days for the weekly forecast fetch.
+const getWeekday = (daysFromToday) => {
+    const now = new Date()
+    const week = ['Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'] 
+    return week[now.getDay() + daysFromToday]
```
magicvilla, line 45: We see what you did there :) another solution is to iterate the array from the api and return weather for each day, and then printing the day e.g: `
filteredForcast.forEach((day)=>{
                const date = new Date (day.dt *1000)
                let dayName = date.toLocaleDateString('en-GB', {weekday:'long'})
`

magicvilla (approved):
Great Job! Your app looks really nice :) It has some interesting features! 
It was a little bit hard to read since you made many changes after the pull request.

Also when you start the app - it would be nice to start from a specifik location and not a blank page. We have left some comments for you - maybe they are not that relevant after your changes. 

GO TEAM TURTLES 🐢 
// Anna & Felicia

script.js
KarolCislowski: Because when you need to use it a lot, and you need to change it, then you need to change it in only one place 🤓

KarolCislowski (approved):
This is a very nice and clean project 🌈 

Remember that you can always come back to it when you have some spare time and add more functionalities to make it an even greater portfolio asset! 🦄 

Great Job!

PR #174, opened 2021-02-21
Project 5 - Tina Jurén & Noemi Frischknecht

NoemiSalome:
Link to our deployed project: [https://project5-tina-noemi.netlify.app/](url)

Working on this project in a pair programming setup was fun and I learned a lot. It's amazing how much further you come when coding in a team than how far i would have gotten all by myself. Still it is hard to make my weekly schedule work since I work during daytime and kind of have a different schedule than most of my team members. My time for this bootcamp is in the evening and on the weekend, but still I want to contribute as much as I can during the week, which is hard and a bit stressfull. I still need some time to find a rhythm that works, but I am getting there. 

Things we'd like to have clarified or explained in more detail:
we asked a question on stackoverflow, but need help with the answer we got. Seems a bit too tricky for us. [https://stackoverflow.com/questions/66300888/how-to-reach-for-the-header-sectionbefore-from-the-js-for-styling](url)

Responsiveness: it is working in the inspect tool, but not when opening it on desktop. We couldn't figure out why and would love to get some hints on this.

/Tina Jurén & Noemi Frischknecht


script.js
```diff
@@ -0,0 +1,71 @@
+const WEATHER_URL = 'https://api.openweathermap.org/data/2.5/forecast?q=Kiruna,Sweden&units=metric&appid=3500e3bacf630605ddf92aeaab386a9e';
```
honkoan, line 1: Does this give also the current weather or only the forecast? Is the current weather displaying right with this url?

honkoan (commented):
Great job with this project Tina and Noemi! 🙌🏻 It was really nice to read your code and see your weather app. I also learned a lot while reading it.

Your code is clean and easy to understand. I also liked your styling a lot! 

I was wondering if the current weather is displaying right. After collaboration days I feel so tired that I may not just understand. 😄 We used different URLs for the current weather and the forecast.

Irina-web-dev (changes requested):
I can only agree with Anna-Kaisa great job with this project! It was not the easy one and considering your time management issues you did it very well and easy to understand. I like the design - it fits Kiruna and Scandinavian style in general 😍 I have two take always from your code - I will definitely use this rounded line someday and defining variables before injecting data into inner HTML is also something I will try to implement in my coding routine as it makes the code look cleaner and easier to understand what data we are actually injecting👌

Big applause to you 👏 and see you soon 🤸‍♀️

P.S. We are not 100% sure, but we think that current information data is not correct, because you are fetching from API endpoint that gives the forecast on the next 5 days and current weather information is not there ⛅ So maybe this is the change to consider before approving the pull request 🕵️‍♀️

script.js
```diff
@@ -0,0 +1,71 @@
...
+const cityName = document.getElementById('city-name');
+const temperature = document.getElementById('temperature');
+const description = document.getElementById('description');
+const container = document.getElementById('forecast-section');
+const todaysImg = document.getElementById('todays-img');
+const temperatureToday = document.getElementById('temperature');
+
+fetch(WEATHER_URL)
```
KarolCislowski, line 9: this is the wrong URL. You need to use the URL specified in STEP1.

KarolCislowski (changes requested):
Please use the correct URL to fetch data to the current weather section.
Please make your page to show days in the forecast sections.
weather-app-tina


When you do it it will be a really great app 🌈

KarolCislowski (changes requested):
Your app is not responsive and you need to have 1 decimal number in your temperatures. 

image

KarolCislowski (approved):
👍🏻

PR #177, opened 2021-02-21
Weather app by Therese and Daniela

dannuzak:
Link to my deployed project:

https://weather-app-therese-daniela.netlify.app

At the beginning of this week, we had a pair programming session where we decided on what approach to use and how we would work. We checked the brief and the requirements and started step by step in order to first reach the blue level, and then also the red. We started with the fetch to display the current weather and then made another fetch for the forecast, where we filtered in order to get the relevant details, and made a loop for the five days. We also decided to fetch the weather icons directly from the API instead of using icons locally saved. After completing the basic requirements we started working on the red level, where we started with changing the color based on weather. We inserted the parts through JS but made the styling in CSS. We also added a fetch for other cities. In order to achieve this we used HTML, CSS and Javascript. We also worked with API and used json files in order to find the correct data to use. We also used media queries in order to make it responsive, and added a footer which is only visible on desktop.

How to work with fetched images a bit more in detail.


code/style.css
```diff
@@ -0,0 +1,199 @@
...
+body.rainy {
+  background-image: linear-gradient(
+    to right top,
+    #637c7b,
+    #718e8c,
+    #7ea09e,
+    #8db2b0,
+    #9bc5c3
+  );
+}
```
esteficodes, line 136: Remember to always keep the same pattern (either use the one you used for: body. clear or the one you used for: body.sunny structure. 👩🏻‍💻

code/script.js
```diff
@@ -0,0 +1,111 @@
...
+  const filteredForecast = data.list.filter(item => item.dt_txt.includes('12:00'))
+
+  filteredForecast.forEach(item => {
+  let temperature = (item.main.temp).toFixed(1);
+  let weekday = (new Date(item.dt * 1000)).toLocaleDateString("en-US", {weekday: "long"});
+  let weatherPicture = `<img src=https://openweathermap.org/img/wn/${item.weather[0].icon}@2x.png></img>`;
+
+  week.innerHTML += `
+    <p>${weekday}${weatherPicture}${temperature} ºC</p>
+    `;
```
esteficodes, line 55: Maybe a little bit coach assistance here, but we think it would be a good idea to put this three elements in different tags to have it easier to make them more responsive?

esteficodes (changes requested):
Good job with this Weather app! It's great that you figured out the way to place the sunset and sunrise times locally, which represented a challenge, given the city you have chosen. Kudos! 💪🏻 //👱🏻‍♀️👩🏻

code/script.js
```diff
@@ -0,0 +1,111 @@
...
+.then((data) => {
+ 
+  let currentWeatherPic = `<img src=https://openweathermap.org/img/wn/${data.weather[0].icon}@2x.png></img>`; 
+  
+  weatherIcon.innerHTML = `${currentWeatherPic}`; 
+  city.innerHTML += ` ${data.name}`; 
+  temperature.innerHTML += ` ${data.main.temp.toFixed(1)} ºC`;
+  weather.innerHTML += `${data.weather[0].description}`;
+  sunrise.innerHTML +=`${new Date((data.sys.sunrise + data.timezone) * 1000).toLocaleTimeString('en-US', {hour: '2-digit', minute: '2-digit', hour12:false,})}`
+  sunset.innerHTML += `${new Date((data.sys.sunset + data.timezone) * 1000).toLocaleTimeString('en-US',{hour: '2-digit', minute:'2-digit', hour12:false,})}`
```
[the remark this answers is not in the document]
KarolCislowski: inside the ${} you can divide your code as you want. The best way is to every part of a chained instructions starts in a new line. 
```
sunrise.innerHTML +=`
  ${new Date((data.sys.sunrise + data.timezone) * 1000)
    .toLocaleTimeString('en-US', {
      hour: '2-digit', 
      minute: '2-digit', 
      hour12:false,
    })}
`
```

KarolCislowski (approved):
This is a very nice project! 🌈  
I like the styling. Is nice and clean!
Estefania left you many nice tips and you should go through them and make your code clean like never 😸 

Great job!!

code/style.css
dannuzak: Here for some reason the Prettier plug in was changing the format automatichally and against my will. 😖

PR #227, opened 2022-09-09
Weather App - Caroline Söderström, Linus Eriksson & Josefine Engdahl

JosefineEngdahl:
Link to project: https://silver-profiterole-7224fd.netlify.app


a-well (approved):
Amazing job on this project! 😀 The code fulfills all the requirements for this project, and works as intended. 

Both Cecilia and I were really impressed with how on point your implementation of the design was, and also what you did with it for tablet and desktop. Super stylish, and functional.
We were also struck by how clean and well written the code was, and by the clever and compact JS. The CSS file also had great structure and we found it easy to follow.

We came across a few extra brackets and some broken indentation/missing spaces, not sure if these were actually breaking anything but figured you might want to know since they looked like the result of merging and/or moving code around, rather than something that was there from the start. I believe resaving the files with a linter installed should fix pretty much all of those for you, so I won't be commenting on them individually (unless you'd like me to) 🙂

Again, really good job! 🌟

/ Amanda Wellington & Cecilia Frånlund

PR #232, opened 2022-09-09
Viktor and Marias Weather app pair project 🥇 

ViktorSvenssonN:
Link to project:
https://sensational-cocada-846c34.netlify.app/

Our reflections on how this project turned out:
We are happy with the results, we had a million ideas for features but we made the wise decision to try to finish and THEN 
add cool features which we noticed that we didn't have time for in the end :) so phew haha :)

Things we like like clarified and explained in more detail:
Was a bit confusing with the pair programming, we understand that we should learn about merges and working
in a dev team. So we did some pair programming in javaScript so both could code the apis and learn :) 

Thank you for this week, we hope who ever reads this have a WONDERFUL day :) 
With kind and warm and sunny regards 
Viktor/Maria

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,4 +9,4 @@ If we had more time, we would love to add a geolocation feature, to change weath
 
 ## View it live
 
-
+https://sensational-cocada-846c34.netlify.app/
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # Weather App
+Viktor and Marias Weather APP :)
 
 A pair project with the goal to develop a weather app. Primary focus to practice JavaScript, learn how to use API and work in branches.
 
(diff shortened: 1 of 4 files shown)


SofiaErikaIngeborg (commented):
Great job with your project this week! 😊
No doubt your code fulfills the requirements, the app works as intended and you have followed the guidelines. 
Your comments are clarly explaining what is happening in each section in the JS.

We found one little tiny bug, for clear weather the link to the icon and the city name doesn't work (very hard to test in this country 😉). We also noticed some small adjustments could be made to the media queries. 

Very clear and well structured code, easy to read. Would recommend to friends and family 💕

Joel & Sofia

PR #234, opened 2022-09-09
Fredrik/Kim Project-weather-app

Earnieball:


Earnieball:
https://fint-vader.netlify.app/

Designs/Design-2/code/app.js
```diff
@@ -0,0 +1,128 @@
...
+  );
+  
+  Promise.all([currentWeather, weatherForecast])
+    .then(responses => {
+      const arrayOfResponses = responses.map(res => res.json());
+      return Promise.all(arrayOfResponses);
+    })
+    .catch(error => {
+      weatherMessage.innerHTML = `<h1 class="error">Something's wrong, try again later.</h1>`;
+    })
```
AgnesAhlman, line 32: Nice with some error handling, but when I try to type in a city that does not exist(or just any rubbish) the error message does not pop up. instead I get the following error message: 

```
app.js:37 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading '0')
    at app.js:37:45
```

AgnesAhlman (commented):
Great job guys! The code is easy to read and looks clean.

Designs/Design-2/code/app.js
karlfredrikholm: Hi Agnes, 

The error message is only showing up if something goes wrong with the fetch, so it's not supposed to show up when you're typing in random stuff. But we agree, it would've been nice with a message when you type in something that it can't find too 😅  

Thanks for your feedback! 🤝

PR #238, opened 2022-09-10
Project-weather-app by Joel and Sofia

joeohm:
https://helpful-pony-3b978e.netlify.app/


script.js
```diff
@@ -0,0 +1,109 @@
...
+
+        sectionForecast.innerHTML += `
+      <div class="forecast-day-container">
+       <p class="forecast-day">${weekday}</p>
+       <p class="forecast-day-temp">${day.main.temp.toFixed(1)}°</p>
+      </div>
+      `;
+      });
+    });
+}
```
lindanorberg, line 108: Nicely done! Short and tidy!

style.css
```diff
@@ -0,0 +1,155 @@
...
+  font-size: 35px;
+  font-weight: bold;
+  padding-bottom: 30px;
+}
+
+.main-container {
+  display: flex;
+  flex-direction: column;
+  width: auto;
+  height: 100%;
```
lindanorberg, line 41: The main cointaner is not covering the full screen. 

Maybe try to add: min-height: 100%;

style.css
```diff
@@ -0,0 +1,155 @@
...
+#164A68 blue text rainy
+#A3DEF7 blue background rainy
+#2A5510 green text sunny
+#F7E9B9 yellow background sunny
+*/
+
+@import url('https://fonts.googleapis.com/css2?family=Montserrat:wght@400;700&display=swap');
+
+body {
+  font-family: 'Montserrat', sans-serif;
```
lindanorberg, line 21: Try to add: height: 100%; 
And maybe target html as well.

style.css
lindanorberg: https://stackoverflow.com/questions/6654958/make-body-have-100-of-the-browser-height

style.css
```diff
@@ -0,0 +1,155 @@
...
+  flex-direction: column;
+  width: auto;
+  height: 100%;
+  padding-bottom: 20px;
+}
+
+/*** City picker section ***/
+
+select {
+  color: unset;
```
lindanorberg, line 48: Thank you for learning us this! We understand that this makes the select forget the previous style, but we don't understand where the select picks up the new style. . .

style.css
```diff
@@ -0,0 +1,155 @@
...
+
+.city-select-clear {
+  background-color: #fff5d5;
+}
+.city-select-clouds {
+  background-color: #ffffff;
+}
+.city-select-rain {
+  background-color: #c3ebfc;
+}
```
lindanorberg, line 78: Nice feature to choose a lighter colour of the select background.

style.css
```diff
@@ -0,0 +1,155 @@
...
+
+.rain {
+  background-color: #a3def7;
+  color: #164a68;
+}
+
+/*** For Tablet ***/
+
+@media (min-width: 768px) and (max-width: 991px) {
```
lindanorberg, line 127: You have the same styling for h1 in both tablet och desktop. If max-width is removed from tablet, the styling goes for desktop as well (exept for the new styling you ad under desktop). And then you dont have to write styling for h1 twice.

lindanorberg (commented):
Nice work! Tidy code! 

We understand the code, the code does not have any duplicates, it works and follows the general requirements. Except for not following the design instructions regarding background color which should stretch al over the browser window.

index.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
```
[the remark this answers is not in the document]
lindanorberg: Oops, now I see that this is part of starting html code. Just ignore this comment. . .

style.css
joeohm: Thanks for the tip and link!
min-height was not needed on main-container in the end when the height and min-height of body/html was set to 100%.

style.css
joeohm: Thanks for the tip, it worked!
```
body,
html {
  font-family: 'Montserrat', sans-serif;
  height: 100%;
  min-height: 100%;
}

style.css
joeohm: Actually min-height was not needed on body/html either, only height: 100%

PR #246, opened 2022-09-11
Week 3- Weather App: Mia Dahlgren & Antonella Cardozo

Sailornina:
https://weather-app-mia-antonella.netlify.app/


Designs/code/script.js
```diff
@@ -0,0 +1,142 @@
...
+
+            const cityValue = json.name;
+            console.log(cityValue);
+            cityDiv.innerHTML += `<h1>${json.name}</h1>`;
+            someText.innerHTML += `<p>In ${json.name} today there <br> are ${json.weather[0].description}</p>`;
+            weatherIcon.innerHTML += ` <img id="main-icon" src="https://openweathermap.org/img/wn/${json.weather[0].icon}@2x.png"  alt="${json.weather[0].description} icon" />`;
+            const weatherDescription = json.weather[0].description;
+            const currentTemp = json.main.temp;
+
+            span1.innerHTML += `${weatherDescription} | ${currentTemp} °c`;
```
NaghmehOkhovat, line 41: We notice that the current temperature is not rounding up or down so maybe you guys can add ...toFixed(0) here too.

Designs/code/script.js
```diff
@@ -0,0 +1,142 @@
...
+                 <span id="temp">${temp}ºC</span>
+                </p>
+                `;
+
+            });
+        });
+
+};
+
+
```
NaghmehOkhovat, line 104: Delete the empty lines between the codes would make it look better.

NaghmehOkhovat (commented):
Great Weather App! 💯 We liked how you guys added comments to your codes because it gives us a better understanding. We only want to point out that it would be better for the code if the naming of certain ids were more descriptive. 
Great job! :)

PR #247, opened 2022-09-11
Weather App by Fanny Cardell & Linnéa Ajger

LinneaAjger:


CeciliaMichele (commented):
Overall a functioning page. Very good naming of the variables which make's it easy to follow to code. When we visit the website during the day, the forecast shows todays date when it's suppose to be showing the weather of tomorrow and the upcoming four days. When we visit the website during the evening, the dates are correct. Is there any way to solve that? 

**HTML**
Very short and clean code.  

**CSS**
The CSS needs to be cleaned up a bit, some attributes are written twice, and there are some unnecessary blanks here and there. 

**Javascript**
Good job using the switch statement! 
There's some console.logs that could be removed for a shorter code. 
You have some divs in your Javascript, the code would be easier to follow if you would have gathered all of them in the HTML.

PR #251, opened 2022-09-11
Weather App | Saralie Bognandi & Julia Östedt

juliaostedt:
View it live: https://stockholm-weather.netlify.app/


code/script.js
```diff
@@ -0,0 +1,107 @@
+// DOM selectors stored as short variables
+const todaysPrompt = document.getElementById("todaysPrompt");
+const header = document.getElementById("header");
+const forecast = document.getElementById("forecast");
+
+// Global variables
```
miadahlgren, line 6: good idea to declare variables this way

code/script.js
```diff
@@ -0,0 +1,107 @@
...
+         <div class="temp"> ${filteredForecast[2].main.temp.toFixed(0)}°</div>
+       </div>
+       <div class="dayTemp">
+         <div class="day">${weekday(filteredForecast[3].dt)}</div>
+         <div class="temp"> ${filteredForecast[3].main.temp.toFixed(0)}°</div>
+       </div>
+       <div class="dayTemp">
+         <div class="day">${weekday(filteredForecast[4].dt)}</div>
+         <div class="temp"> ${filteredForecast[4].main.temp.toFixed(0)}°</div>
+       </div>
```
miadahlgren, line 103: this part is a bit repetitive so a loop would be nice. But we know that you would have added that if you had more time. :) Is there another way to do it?

code/style.css
```diff
@@ -0,0 +1,70 @@
...
+.dayTemp {
+  display: flex;
+  justify-content: space-between;
+  padding: 5px;
+  width: 90%;
+  border-bottom: dashed 1px;
+  text-transform: lowercase;
+}
+
+/* tablet styles */
```
miadahlgren, line 39: the responsiveness works really well! We have nothing bad to say :D

miadahlgren (commented):
you did a really good job! if you want to continue working on the project you could change the  const weekday (liness 83-103) into a loop of some sort but apart from that the app looks really great. The code was well structured. //Mia and Antonella

Sailornina (commented):
Good Job! :)

PR #252, opened 2022-09-11
Weather-app Nina Berggren, Linda Norberg

NinaBerggren:


code/script.js
```diff
@@ -0,0 +1,125 @@
...
+        }
+        
+        todaysWeather.innerHTML = `
+          <p>
+            ${(weatherData.weather[0].description).toLowerCase()} | ${Math.round(weatherData.main.temp * 10) / 10}°
+          </p>
+          <p>
+            sunrise ${new Date(weatherData.sys.sunrise * 1000).toLocaleTimeString([], { hour: '2-digit', minute: '2-digit'})}
+          </p>
+          <p>
```
Archkrull, line 52: Maybe we could make the "toLocaleTimeString" into one function rather than writing the same thing twice?

Archkrull (commented):
Overall, the code works and meet all of the blue requirements! It looks really good and follows the design.

We found the code very clean, easy to understand and follow. The comments are also really good and descriptive.
We only had one comment, to make some code into one function rather than two. Overall, this is a great app!

code/script.js
lindanorberg: True! Thanks for the suggestion!

PR #276, opened 2023-02-25
Jennifer and Joanna's weather app 

jenster5:
https://weather-app-technigo-jen.netlify.app/


EvaSjoholm (commented):
I don't have any remarks as well as HTML and JS, I could easily follow your code and its functions for the webpage, well done! 
For the CSS maybe I'd consider changing in the section weather-today and "feels like", so it fits it's content. Page was overall really good on both Safari and Chrome.

PR #277, opened 2023-02-25
W3-project made together with Emilia G. Anundi

YlvaKarlsson:


cecavi (commented):
Nice job guys, over all very short and clean code.  Easy to read and good of you to write comments to each function. 

We liked your choices of colors and we see that your css file has the most code. Nice job with the styling!

You meet the requirements for blue level but you need to adjust the responsivness for tablet and desktop. It might be as easy as changing the value of width from fixed pixels  to percentage. 

/Cecilia Avila and Nina Ströyer

PR #285, opened 2023-02-26
Weather app by Camilla Cronqvist & Sofia Gerdmar

camcron:
The JavaScript code is working, but unfortunately not DRY.


jonsjak (commented):
Good looking design with nice gradient, though very different from the original (design 1?) which makes it hard to compare.

The weather condition and temperature is a bit hard to see when underlined with dots.

The code gets a bit repatative because it runs the same function for each city displayed on the page. If you use a changeable URL like: https//:getweatherfor${city}.com and just change let the user change city (with a let city = userInput.value) you could make a general function for whatever city user inputs and only display one city at a time.

Otherwise everything is working as it should. Good job!

PR #288, opened 2023-02-26
Weather app by Giorgio Cugusi and Regina Söderberg

GiorgioCugusi:


index.html
```diff
@@ -0,0 +1,51 @@
...
+      <!--Header--> 
+        <div id="header" class="header"></div>
+        <div id="baseContainer" class="base-container">
+        <!--Today's prompt-->
+       
+        <!--Forecast-->
+        <div id="forecast" class="forecast">
+<container class="report">
+
+       <!--  <div class="days">
```
AnnikaSonnek, line 29: Seems like you also had to comment out the names for each day? We had trouble that it wouldn't replace what was in our forecast container, instead, the values from the API would only place themselves below instead of replacing the already existing content. Did you have the same issue?

script.js
```diff
@@ -0,0 +1,103 @@
+// DOM selectors stored as short variables
+const todaysPrompt = document.getElementById("todaysPrompt");
+const header = document.getElementById("header");
+const forecast = document.getElementById("forecast");
+
+// Global variables
+let Url_Weather =
```
AnnikaSonnek, line 7: Nice to place the API url in a variable instead of having it in the fetch!

script.js
```diff
@@ -0,0 +1,103 @@
...
+        //Declare variables for the time of sunset and sunrise. new Date () changes the UNIX time to day/date/year/hh:mm:ss/time zone.
+        const sunrise = new Date(json.sys.sunrise * 1000).toLocaleTimeString([], {
+            hour: "2-digit",
+            minute: "2-digit",
+          });
+          const sunset = new Date(json.sys.sunset * 1000).toLocaleTimeString([], {
+            hour: "2-digit",
+            minute: "2-digit",
+          });
+            //Convert timestamp to milliseconds with *1000. JavaScript stores Dates in milliseconds.
```
AnnikaSonnek, line 32: Nice with comments on this one!

script.js
```diff
@@ -0,0 +1,103 @@
...
+          
+
+          header.innerHTML += `${todaysWeather} | ${todaysTemperature}°<br>sunrise ${sunrise}<br> sunset ${sunset}`;
+
+          if (todaysMainWeather === "Clear") {
+            todaysPrompt.innerHTML = `<div class= "Clear">
+                      <img class= "sunny" src="./Designs/Design-1/assets/Group36.png"/>
+                      <h2> Get your sunnies on. ${json.name} is looking rather great today.</h2>
+                      </div>`;
+            document.body.style.backgroundColor = "#F7E9B9";
```
AnnikaSonnek, line 44: Oh cool! So the background would change depending on the weather?

style.css
```diff
@@ -0,0 +1,124 @@
...
+
+.day {
+  color: #4d4c4c;
+}
+
+.temp {
+  color: #4d4c4c;
+}
+
+@media  (max-width: 768px) {
```
AnnikaSonnek, line 114: When you go below 937 px in width, there is a yellow background that appears after the background image. I think it could be solved by repeating the image in CSS.

script.js
```diff
@@ -0,0 +1,103 @@
...
+  })
+  .then((json) => {
+    // new variable to filter the table and choose the same time everyday.
+    const filteredForecast = json.list.filter((item) =>
+      item.dt_txt.includes("12:00")
+    );
+    console.log(filteredForecast);
+    forecast.innerHTML += `
+       <div class="forecast">
+         <div class="day">${weekday(filteredForecast[0].dt)}</div>
```
AnnikaSonnek, line 83: If you don't want to write out each day specifically (or if you hade more values to display), you could use the foreach-loop to loop through the array filteredForecast. Same thing with the temperatures for each day.

script.js
AnnikaSonnek: I just now saw the images you used. They were really nice! ^_^

AnnikaSonnek (commented):
Good job with the project! The only thing I would add is maybe more comments and explain what the code does.

PR #308, opened 2023-09-15
Weather-app-Mina & Janice

Minadarabi:
Deployed on Netlify : https://master--resplendent-basbousa-b3cbf0.netlify.app/


code/index.js
```diff
@@ -0,0 +1,271 @@
...
+for (let index = 1; index < 6; index++) {
+  const date = new Date(today);
+
+  date.setDate(today.getDate() + index);
+  const dayName = daysName[date.getDay()];
+  nextFiveDAys.push(dayName)
+
+}
+
+daysForecast.forEach((dayForc, index) => {
```
mariateresepettersson, line 148: Cool, have not seen this before. What does the substract do in this case?

code/index.js
```diff
@@ -0,0 +1,271 @@
...
+  tempElement.innerText = `${Math.floor(data.main.temp)}`;
+  windMain.innerText = `${Math.floor(data.wind.speed)} m/s`;
+  cityName.innerText = data.name;
+  descriptionEl.innerText = data.weather[0].description;
+  iconMain.src = `https://openweathermap.org/img/wn/${data.weather[0].icon}@2x.png`;
+
+  // time now
+
+  let newTime = new Date((data.dt + timeZone + (new Date().getTimezoneOffset() * 60)) * 1000).toLocaleTimeString([], { timeStyle: 'short', hour12: false });
+  console.log(newTime)
```
mariateresepettersson, line 217: Great to see that you have used console.log to check your code along the progress. Perhaps it would look cleaner if you commented it out or took it away when the code is ready :)

code/index.js
```diff
@@ -0,0 +1,271 @@
...
+
+  timeToday.innerText = newTime;
+
+  // images 
+
+  const weatherInfo = data.weather[0].main;
+  const isDaytime = newTime >= newSunrise && newTime < newSunset;
+  let backgroundImage;
+
+
```
mariateresepettersson, line 229: Nice to see how clean your code looks because you decided to put the images as objects to recall in the function later on! Great job!

code/index.html
```diff
@@ -0,0 +1,111 @@
...
+                <img src="" alt="" class="icon">
+                <p class="temp-forecast"></p>
+                <p class="wind-forecast"></p>
+            </div>
+            <div id="day5" class="days">
+                <p class="day"></p>
+                <img src="" alt="" class="icon">
+                <p class="temp-forecast"></p>
+                <p class="wind-forecast"></p>
+            </div>
```
mariateresepettersson, line 103: We understand that this is a fixed structure, but maybe it would be possible to add this in the JS to manipulate the DOM and keep the html cleaner?

mariateresepettersson (commented):
Super impressive work! We mostly have some suggestions and some question regarding clarification. Otherwise great structure, easy to follow your code, good naming of variables and functions. Nice design!

code/index.js
Minadarabi: in this case i wanted to take a day name just 3 of letters for example Monday => Mon so i have used substr function (0,3) just have 3 letters of the day name

code/index.js
Minadarabi: Yes, you're right , it will be more clear to read

code/index.html
Minadarabi: great view , thank you for the recommendation , yes if we took it in JS, it would be nicer

PR #313, opened 2023-09-16
Sandra and Mannes weatherapp

SVMaxica:
https://sandraandmannesweatherapp.netlify.app/


code/index.html
```diff
@@ -0,0 +1,53 @@
...
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Weather App</title>
+    <link rel="stylesheet" href="style.css">
+</head>
+<body>
+    <div class="app">
+        <header>
+           
+        </header>
```
andrkavr, line 13: Can't see that this is used for anything. It is styled in the css file, but appears to have no functional use.

code/script.js
```diff
@@ -0,0 +1,199 @@
...
+        </div>
+      </div>`;
+      });
+    })
+    .catch((error) => {
+      console.error('Error:', error);
+    });
+};
+fetchCurrentWeather();
+  fetchFiveDayWeather();
```
andrkavr, line 165: Maybe insert these into a function called start or add a comment that informs that this is where script is run for the first time. Should probably be moved to the bottom of the script page to separate from other logic.

Isaackamran (commented):
The coding looks clean and easy to follow. A few things that I noticed in the deploy was if I click enter more than once the forecast disappears and I think it has somthing to do with the script.js row 173 and 174. Also the icons you have added is emojis if im not mistaken, could be intresting to fetch the given icons from the API.

I can see that you worked on the design of the weather app with a header and container, however it doesn't look as any of the examples we were given. Maybe it was lack of time. Nice search bar! Good job!

code/script.js
```diff
@@ -0,0 +1,199 @@
...
+  } else if (weatherID >= 700 && weatherID < 800) {
+    return "🌫️";
+  } else if (weatherID === 800) {
+    return "☀️";
+  } else if (weatherID >= 801 && weatherID < 900) {
+    return "☁️";
+  } else {
+    return "❌";
+  }
+}
```
andrkavr, line 97: Nicely done using the id's! It's nice to see how we approach the same challenge in different ways. We went with using Main in our project.

code/script.js
```diff
@@ -0,0 +1,199 @@
...
+          <span class="day-forecast">${day}</span>
+          <span class="temperature-forecast">${temperature}°C</span>
+        </div>
+      </div>`;
+      });
+    })
+    .catch((error) => {
+      console.error('Error:', error);
+    });
+};
```
andrkavr, line 163: This is not displayed if a user submits a search without input or if the city is misspelled.

code/style.css
```diff
@@ -0,0 +1,170 @@
...
+    text-align: center;
+    padding: 20px;
+}
+
+/* Style the header */
+header {
+   /*background-color: #3498db;*/
+    color: white;
+    padding: 0px;
+}
```
andrkavr, line 24: Could be removed if the header isn't displayed.

code/style.css
```diff
@@ -0,0 +1,170 @@
...
+
+h2{
+    font-size: 38px;
+    padding-top: 20px;
+}
+
+.forecast-container {
+
+
+}
```
andrkavr, line 117: Doesn't style anything.

andrkavr (approved):
All in all a well executed project that looks great.
We could see that you took some inspiration from one of the designs, but mostly did your own thing with it. 

Nicely done!

PR #315, opened 2023-09-17
Wk4 - Weather App - Stenli and Beckie

BeckieMorton:


Soygirt (approved):
You've made a great site even though you've had problems finding a time to code together because of schedules and timezones! We loved the rain effect! 

The code looks clean with great comments that makes it easy to find the different sections in your code.
Great use of json, and the 5day forecast looks good!

PR #318, opened 2023-09-17
Weather app

AmandaRam:


VAstrom (approved):
Great job! I didn't notice any bugs.
I think the page looks great and very close to the design provided (the only difference being the search bar) so I think that is really impressive! 
The code over all is very neat and DRY.  
I really liked the file with the steps and I'll try to do something similar in my projects since it gives a very clear picture of what needs to be done.

MikoZhu (commented):
Thank you for sharing your code. I found your use of try...catch, the loading section, and padStart methods to be excellent, reflecting both good code practices and sound logic. I've learned a lot from your code. However, I also have a few minor suggestions, which I've listed below. Well done👍

PR #323, opened 2023-09-17
Project-weather-app

malinlunde:


Hang-Nguyen-Vu (commented):
It is a pity that geolocation is not working. I have no experience with it, so unfortunately cannot help. Apart from that, the page looks neat and responsive. The data is presented in correct format. Love the weather animation! For the forecast, I would prefer replacing some information/text with icons, e.g. weather description, or removing those whose data is self-explanatory, e.g. temperature. Another thought is that while geolocation allows the navigator to get weather data based on their longitude and latitude, that person may still want to search for data about another city. In that sense, it would be great to still have a search bar. It seems complicated to have both fetching data by city name and by geolocation, so just a thought.

PR #328, opened 2023-09-17
project weather app - andrkavr & Isaackamran

andrkavr:


KroLuna (approved):
Great work Andrej and Isaak! The app works nicely and the requirements are fulfilled. The code looks good and it's readable. The lay out looks as the design we where given in the project. Maybe a future goal could be to put the round button to change cities. Once again great work! Caro/Frida 🥳

PR #333, opened 2023-09-19
weather app

hamdiolad:
weather app by hamdi olad and evelyn del carmen.


code/script.js
```diff
@@ -0,0 +1,130 @@
...
+
+                if (!dailyForecast[day]) {
+                    dailyForecast[day] = {
+                        temperature: forecastData.main.temp,
+                        description: forecastData.weather[0].description,
+                    };
+                } else {
+
+                    dailyForecast[day].temperature = forecastData.main.temp;
+                    dailyForecast[day].description = forecastData.weather[0].description;
```
AmandaRam, line 103: Description is never used in the forecast so it can be removed.

code/script.js
```diff
@@ -0,0 +1,130 @@
...
+            const dailyForecast = {};
+
+            forecast.list.forEach((forecastData) => {
+                const date = new Date(forecastData.dt * 1000); // Convert Unix timestamp to milliseconds
+                const day = date.toLocaleDateString(undefined, { weekday: 'long' });
+
+
+                if (!dailyForecast[day]) {
+                    dailyForecast[day] = {
+                        temperature: forecastData.main.temp,
```
AmandaRam, line 97: toFixed(1) should be added to get the temperature with only one decimal point.

code/script.js
```diff
@@ -0,0 +1,130 @@
...
+        .then((forecast) => {
+            weeklyforecast.innerHTML = '';
+
+
+            const dailyForecast = {};
+
+            forecast.list.forEach((forecastData) => {
+                const date = new Date(forecastData.dt * 1000); // Convert Unix timestamp to milliseconds
+                const day = date.toLocaleDateString(undefined, { weekday: 'long' });
+
```
AmandaRam, line 93: Filter out every date that is today. The current weather should only be displayed in the header

code/script.js
```diff
@@ -0,0 +1,130 @@
...
+                const temperatureCell = document.createElement('td');
+
+
+                dayCell.textContent = day;
+                temperatureCell.textContent = `${dailyForecast[day].temperature} °C`;
+
+                row.appendChild(dayCell);
+                row.appendChild(temperatureCell);
+                weeklyforecast.appendChild(row);
+            }
```
AmandaRam, line 120: could be simplified as:
```
weeklyforecast.innerHTML += `
    <tr>
        <td>${day}</td>
        <td>${dailyForecast[day].temperature} °C</td>
    </tr>
`;
```

AmandaRam: You could also, instead of using a table do something like this:
```
<div class="forecast">
    <p>${day}</p>
    <p>${dailyForecast[day].temperature} °C</p>
</div>

...

// In CSS

.forecast {
    display: flex;
    justify-content: space-between;
}
```

AmandaRam (changes requested):
Nice job! 🏅  Overall the code is working, but there are some details that could be fixed in order to follow the design and also to keep the code more clear. A friendly reminder - pls add some comments for each step i the code to make it easier to understand and follow. 💯

PR #364, opened 2024-03-01
Annikas Weather app

annsig:
## Netlify link
 https://relaxed-fudge-449ac4.netlify.app/


script.js
```diff
@@ -0,0 +1,62 @@
...
+        const filteredArray = forecastArray.filter((day) => {
+            return day.dt_txt.toLowerCase().endsWith("12:00:00");
+        });
+
+        const forecastContainer = document.querySelector('.forecast__container');
+        filteredArray.forEach((day, index) => {
+            if (index < 4) {
+                let timestamp = day.dt;
+                let date = new Date(timestamp * 1000);
+                let dayOfTheWeek = date.toLocaleDateString("en-US", { weekday: "short" });
```
HIPPIEKICK, line 48: I think these can be constants

script.js
```diff
@@ -0,0 +1,62 @@
...
+                forecastContainer.innerHTML += `
+                    <div class="forecast__single-day-flex">
+                        <p class="forecast__day">${dayOfTheWeek} - ${parseInt(day.main.temp)}°C ${day.wind.speed}</p>
```
HIPPIEKICK, line 51: Maybe you could add some space and a symbol to show this is windspeed?

styles.css
```diff
@@ -0,0 +1,69 @@
...
+.multiline-text {
+    white-space: pre-wrap;
+    display: inline;
+}
+
+.padding{
+    border: 1px #0000;
+    border-radius: 10%;
+    width: 375px;
+    height: 812px;
```
HIPPIEKICK, line 66: What's the purpose of this? It's causing a side scroll on devices that are more narrow than 375px

HIPPIEKICK (changes requested):
Good job with your weather app Annika! Some small things still need fixing:
- Make your app responsive (it should look good on devices from 320px width up to 1600px)
- The presentation of the data should be in the specified format (1 decimal on the temp)

PS. Also, make sure you have your VSC indentation set to 2 spaces (right now it looks like you have 4).

PR #367, opened 2024-03-01
project-weather-app Viktoria O.

viktoria-olo:
## Netlify link
https://marvelous-empanada-370879.netlify.app/

## Collaborators
Individual project


index.html
```diff
@@ -0,0 +1,60 @@
...
+    <link
+      href="https://fonts.googleapis.com/css?family=Montserrat"
+      rel="stylesheet"
+    />
+    <title>Weather Compass</title>
+  </head>
+  <body id="body">
+    <section id="weather-info-container">
+      <div>
+        <table class="weather-today">
```
jyy009, line 16: Cool that you used tables to structure it as I wouldn't have thought to do that!

script.js
```diff
@@ -0,0 +1,242 @@
...
+    );
+    // Create an array with the temperatures of the filtered objects
+    const dailyTemperatures = forecastArray.map((degrees) => degrees.main.temp);
+    // Create an array with the timestamps of the filtered objects
+    const timestampArray = forecastArray.map((times) => times.dt);
+    timestampArray.forEach((timestamp, index) => {
+      const dayDate = new Date(timestamp * 1000);
+      weekday = dayDate.getDay();
+      // Check so that the forecast for the current day is not included
+      if (weekday !== currentDay) {
```
jyy009, line 191: Great that you figured how to compare the indexes to match the day with the temp!

jyy009 (commented):
This looks awesome! It's super well organized, structured and was really easy to follow along. Also your variable names were informative which helped the follow along process. And I liked how your image is big on the right side when on desktop view.

script.js
```diff
@@ -0,0 +1,242 @@
...
+/////////////////////////// Global Variables ////////////////////////////
+
+const BASE_URL = "https://api.openweathermap.org/data/2.5/";
+const API_KEY = "5f9b0149e7c813c77ba22f081321a0c1";
+const weatherData = "weather";
+const forecastData = "forecast";
+let locale = "Stockholm,Sweden";
+const units = "metric";
+let URL = `${BASE_URL}${weatherData}?q=${locale}&units=${units}&APPID=${API_KEY}`;
+let forecastURL = `${BASE_URL}${forecastData}?q=${locale}&units=${units}&APPID=${API_KEY}`;
```
HIPPIEKICK, line 23: These could be constants I think?

script.js
```diff
@@ -0,0 +1,242 @@
...
+  cityName.innerHTML = `Get your sunnies on. ${localCity} is looking rather great today.`;
+};
+
+// Function that changes the design for rainy days
+const setRainDesign = () => {
+  smallWeatherIcon.src = "./design/design2/icons/noun_Umbrella_2030530.svg";
+  largeWeatherIcon.src = "./design/design2/icons/noun_Umbrella_2030530.svg";
+  body.className = "rain";
+  cityName.innerHTML = `Don't forget your umbrella. It's wet in ${localCity} today.`;
+};
```
HIPPIEKICK, line 61: Could these be one function but with if/else statements instead base on a param? 👀 e.g.
```
const updateDesign = weather => {
  if (weather === "Rain") {
    //Logic
  }
}
```

script.js
```diff
@@ -0,0 +1,242 @@
...
+      weekday = "fri";
+      break;
+    case 6:
+      weekday = "sat";
+      break;
+    default:
+      weekday = "unknown";
+      break;
+  }
+};
```
HIPPIEKICK, line 174: Nice with a switch! Another option is to make use of the indexes

script.js
HIPPIEKICK: Plus one on that 👍

index.html
HIPPIEKICK: Same ☝️

HIPPIEKICK (changes requested):
Interesting approach you've taken! It's working, so I won't ask you to change it - but I want to address that you're making multiple API calls to fetch the weather, city name, temperature, and sunrise/sunset data. It's repetitive (and we should try to follow the DRY principle). Next time, consider fetching all required data in a single API call and storing it in variables (or state in React which we will start talking about next week) for reuse.

What I _do_ want to ask you to change though is the layout on mobile. On 320px device width we get a tiny side scroll caused by the first cell in the table being a bit too wide. From requirements:
- Make your app responsive (it should look good on devices from 320px width up to 1600px)

Apart from that - it looks good and you're code is structured ⭐

HIPPIEKICK (approved):

PR #368, opened 2024-03-01
Ericas weather-app

ericamechler:
## Netlify link
https://ericas-weather-app.netlify.app/


fetch-weather-api.js
```diff
@@ -0,0 +1,217 @@
...
+    // Display moon-image in html
+    todaysTimeOfDay.innerHTML += `<img src="./design/assets/Big-moon.png" alt="Big moon image">`;
+    // Add a class for night mode to the weatherBackground
+    weatherBackground.classList.add("night-mode");
+  }
+};
+
+// The fetching-function
+const fetchWeatherToday = (url) => {
+  fetch(url)
```
vidalhuix, line 123: I'm not sure, but should we put the url here as a parameter? I asked GPT and says that fetch in line 123 should have a return before to properly handle asynchronous operation.
```javascript
  return fetch(url)
```

index.html
```diff
@@ -0,0 +1,56 @@
...
+      href="https://fonts.googleapis.com/css2?family=Handlee&family=Nunito+Sans:opsz,wght@6..12,200;6..12,300&family=Nunito:wght@200&family=Roboto:ital,wght@0,300;1,300&display=swap"
+      rel="stylesheet"
+      
+    />
+    <title>Ericas weather app</title>
+  </head>
+  <body>
+    <div class="weather-background" id="weather-background">
+    <header>
+      <!-- hamburgermenu for show -->
```
vidalhuix, line 23: We could have used `<nav>` here instead of `<header>`.

index.html
```diff
@@ -0,0 +1,56 @@
...
+            <div class="todays-sunrise" id="sunrise"></div>
+            <div class="todays-sunset" id="sunset"></div>
+          </div>
+        </div>
+        <!-- div with the time of day for icons -->
+        <div class="todays-time-of-day" id="todays-time-of-day">
+        </div>
+    </div>
+    <button class="switch-button" id="switch-button">
+        <img src="./design/assets/arrow-up-3100.png">
```
vidalhuix, line 46: Consider adding an empty alt="" attribure to improve accessibility.

index.html
```diff
@@ -0,0 +1,56 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no" />
+    <link rel="stylesheet" href="style.css" />
+    <script
+      src="https://kit.fontawesome.com/2e83762929.js"
```
vidalhuix, line 8: Do we need that script here?

style.css
```diff
@@ -0,0 +1,158 @@
+* {
+  box-sizing: border-box;
+  margin: 0px;
+  padding: 0px;
+}
+
+:root {
```
vidalhuix, line 7: Nice with the :root here!

style.css
```diff
@@ -0,0 +1,158 @@
...
+  --font-size-xs: 12px;
+  --font-size-s: 14px;
+  --font-size-m: 20px;
+  --font-size-l: 30px;
```
vidalhuix, line 11: Font Units: When specifying font sizes, it's generally recommended to use relative units like em, rem, or percentages (%) instead of fixed units like px. This allows for better scalability and accessibility, especially for users who may have their browser font sizes adjusted.

style.css
```diff
@@ -0,0 +1,158 @@
...
+}
+
+.switch-button {
+  position: absolute;
+  z-index: 2;
+  font-size: var(--font-size-s);
+  background-color: #757aff;
+  padding: 15px;
+  color: white;
+  border-radius: 100%;
```
vidalhuix, line 126: Border on Switch Button: The border property in the .switch-button class is missing a border style (e.g., solid, dotted, dashed). You may want to specify a border style to ensure consistent rendering across different browsers.

style.css
```diff
@@ -0,0 +1,158 @@
...
+  justify-content: space-between;
+  font-size: var(--font-size-xs);
+  align-items: center;
+  width: 100%;
+  text-align: center;
+}
+
+.forecast-day p {
+  width: 100%;
+}
```
vidalhuix, line 158: Media queries?

vidalhuix (commented):
Nice code, easy to read and easy to follow with the comments:)
Great jobb Erica! 

PS: I have put some of the comments directly from GPT, so u know :P

vidalhuix:
Ah! I forgot to say to delete all the unnecessary folders in the repo:)

fetch-weather-api.js
```diff
@@ -0,0 +1,217 @@
...
+let city = "Lundsberg";
+const API_KEY = "ebcad7517d4d5102daa2078b4d1b8409";
+let weatherTodayURL = `${BASE_TODAY_URL}q=${city}&units=metric&APPID=${API_KEY}`;
+
+// Function to update UI with todays weather
+const updateWeatherToday = (weatherTodayData) => {
+  /* ............START OF created variables........... */
+  if (weatherTodayData.cod === "404") {
+    // If city is not found, display an alert
+    alert("Sorry, the city was not found. Please try again.");
```
HIPPIEKICK, line 23: Nice error handling 👍

fetch-weather-api.js
```diff
@@ -0,0 +1,217 @@
...
+  const timezoneOffsetMilliseconds = localTime.getTimezoneOffset() * 60000; // Convert minutes to milliseconds
+  const localTimeCorrected = new Date(
+    localTime.getTime() + timezoneOffsetMilliseconds
+  );
+
+  //Format the corrected local time in Swedish locale
+  const formattedLocalTime = localTimeCorrected.toLocaleTimeString("sv-SE", {
+    hour: "2-digit",
+    minute: "2-digit",
+  });
```
HIPPIEKICK, line 80: But what if I want to see the the weather in another timezone? 👀

fetch-weather-api.js
HIPPIEKICK: The return happens in the .then blocks, so all is good. Maybe you reacted to the code being split on different lines?

fetch-weather-api.js
```diff
@@ -0,0 +1,217 @@
...
+    })
+    .catch((error) => {
+      console.error("Error fetching today's weather", error);
+    });
+};
+
+// Creating API URL called weatherForecastURL
+const BASE_FORECAST_URL = "https://api.openweathermap.org/data/2.5/forecast?";
+
+let weatherForecastURL = `${BASE_FORECAST_URL}q=${city}&units=metric&APPID=${API_KEY}`;
```
HIPPIEKICK, line 138: Why is this a let?

fetch-weather-api.js
```diff
@@ -0,0 +1,217 @@
...
+    // Get temperature
+    const temperature = filteredData.main.temp.toFixed(0);
+
+    forecastList.innerHTML += `
+    <div class="forecast-day">
+    <p> ${dayOfWeek} </p>
+    <img src="${iconUrl}" alt="Weather Icon">
+    <p> ${temperature} °C</p>
+    </div>
+    `;
```
HIPPIEKICK, line 174: Don't forget indentation:
```
forecastList.innerHTML += `
  <div class="forecast-day">
    <p> ${dayOfWeek} </p>
    <img src="${iconUrl}" alt="Weather Icon">
    <p> ${temperature} °C</p>
  </div>
`;
```

fetch-weather-api.js
```diff
@@ -0,0 +1,217 @@
...
+      updateWeatherForecast(weatherForecastData);
+    })
+    .catch((error) => {
+      console.error("Error fetching weather forecast:", error);
+    });
+};
+
+// Trying function for the search-bar
+const loadSearchResults = (event) => {
+  if (event.key === "Enter") {
```
HIPPIEKICK, line 191: Nice!

style.css
```diff
@@ -0,0 +1,158 @@
...
+
+  --background-color-standard: #f5f5f5;
+}
+
+body {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  margin: 30px;
```
HIPPIEKICK, line 28: When screens are narrow, it's nice to make use of the whole screen's width so consider removing the margin

style.css
HIPPIEKICK: I would recommend to increase the font-size a bit. 12px is hard to read 🔍

HIPPIEKICK (approved):
Good job with the weather app! 🌞 

Apart from the comments I wrote, a thing I noticed is that the background colour doesn't go back to light after it has been dark (probably because of CSS being cascading). Consider always removing the other class name before adding another.

fetch-weather-api.js
ericamechler: Seemed like it had to be a let to be able to get the search-function working(to be able to updated the API URLs with the new city)

PR #372, opened 2024-03-02
Weather app Vittoria

vittoriamatteoli:
## Netlify link
https://weatherappbyvittoria.netlify.app/
## Collaborators
[vittoriamatteoli]


style.css
```diff
@@ -0,0 +1,198 @@
...
+    flex-direction: row;
+  }
+  audio {
+    width: 600px;
+  }
+}
+
+/* snow animation */
+
+.snow-day {
```
kathinka, line 119: maybe we should store the vector files in a separate document we can reference here to make the style document easier to read? and avoid typo errors?

style.css
```diff
@@ -0,0 +1,198 @@
...
+  -o-background-size: 450px auto;
+  background-size: 450px auto;
+  -webkit-animation: snowing 20s linear infinite;
+  -moz-animation: snowing 20s linear infinite;
+  -ms-animation: snowing 20s linear infinite;
+  -o-animation: snowing 20s linear infinite;
+  animation: snowing 20s linear infinite;
+}
+
+@-webkit-keyframes snowing {
```
kathinka, line 137: very cool to see the use of keyframes animation 👯

kathinka (commented):
Hi Vittoria!

Wow you should be so proud 🥇 
How fun it was to be reviewing you amazing project! you have done so well. you have fullfilled the assignement and went even beyond with animation, search and sound! It looks great on DT and mobile in safari, ff and chrome!

big applause! I struggled to find something to mention, but I made some suggestions.

one last tip:

maybe you could consider making the search area more intutive for the user, (for instance; maybe just a border around the input field would suffice?) 

But thats just me trying to find something to improve. Again, you should be really proud! Great work!

script.js
```diff
@@ -0,0 +1,314 @@
...
+const navContainer = document.getElementById('navContainer')
+const thunderAudio = document.getElementById('thunderAudio')
+const rainAudio = document.getElementById('rainAudio')
+const sunnyAudio = document.getElementById('sunnyAudio')
+//variables
+let currentDate
+let currentDay
+let currentTime
+let API_URL = `https://api.openweathermap.org`
+let API_KEY = `1c745605f5cf52ece2c729289e47acc7`
```
HIPPIEKICK, line 14: These can be constants and you can use '' instead of ``

script.js
```diff
@@ -0,0 +1,314 @@
...
+      weatherConditions.forEach((condition) => {
+        switch (true) {
+          // HTML for thunderstorms
+          case condition >= 200 && condition <= 232:
+            document.body.style.backgroundColor = '#212f54'
+            document.body.style.color = '#FFD43B'
+            controls.style.display = 'none'
+            document.getElementById('thunderAudio').controls = true
+            document.getElementById('rainAudio').controls = false
+            document.getElementById('sunnyAudio').controls = false
```
HIPPIEKICK, line 87: Since you created these in the HTML file, these could be global DOM selectors to keep it DRY

script.js
```diff
@@ -0,0 +1,314 @@
...
+    getWeatherCities(city)
+    cityInput.value = ''
+  } else {
+    console.log('Please enter a city name.')
+  }
+}
+
+// event listener
+const searchButton = document.querySelector('button')
+searchButton.addEventListener('click', handleCityInput)
```
HIPPIEKICK, line 287: Would be nice to also listen to the pressing of an Enter button

script.js
```diff
@@ -0,0 +1,314 @@
...
+}
+
+// function to geolocate
+const getLocation = () => {
+  if (navigator.geolocation) {
+    navigator.geolocation.getCurrentPosition(showPosition)
+  } else {
+    body.innerHTML = 'Geolocation is not supported by this browser.'
+  }
+}
```
HIPPIEKICK, line 309: This function is never called

HIPPIEKICK (approved):
So Zürich is rainy today, huh? ☔ A year ago, I lived on Rotwandstrasse and had a lot of spring feelings. I remember a church close to the apartment where there was a flowerbush that smelled AMAZING.

Well well, good job with your weather app Vittoria! Nice to see you play around with the stretch goals as well 🌞

script.js
vittoriamatteoli: Thank you! I must have forgotten to call it afterwards 😅

script.js
vittoriamatteoli: I tried to make them global but in that case I would have the controls buttons showing immediately after the page was loading.

style.css
vittoriamatteoli: yes good idea! thank you

script.js
```diff
@@ -0,0 +1,314 @@
...
+  // prevent audio to be played
+  rainAudio.pause()
+  sunnyAudio.pause()
+  thunderAudio.pause()
+  // validation for cities
+  if (/^[a-zA-Z\s-]+$/.test(city)) {
+    getWeatherCities(city)
+    cityInput.value = ''
+  } else {
+    console.log('Please enter a city name.')
```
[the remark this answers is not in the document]
vittoriamatteoli: Yes good point ,thanks

PR #402, opened 2024-03-04
Weather condition logo

jingh999:
## Netlify link
Add your Netlify link here.
PS. Don't forget to add it in your readme as well.

## Collaborators
Add your collaborators here. Write their GitHub usernames in square brackets. If there's more than one, separate them with a comma, like this:
[github-username-1, github-username-2]


jingh999:
Wrong base.

PR #414, opened 2024-09-27
Xings Weather App

xingyin2024:
https://xingsweatherapp.netlify.app/


HIPPIEKICK (dismissed):
**HTML/CSS**
- Well structured HTML, with semantic elements 👍 
- Clean and organized CSS file as well with nice usage of CSS variables.

**JavaScript**
- Nice structure overall as well as well-named variables
- Impressive work with the stretch goal to show more cities!
- Nice error handling!
- One of the learning goals is understanding how to use JavaScript’s Date object. While you’ve fetched the current weather and the local time, it would make sense to display the local date and time for the **searched location**. I find it confusing that it's my local time for the cities that I know are in a different time zone than I am. This is true for both the Current Time and the sunset/sunrise times.
- Really good that you divided the code into modular functions, makes it more maintanable and easy to follow!

**Changes Requested**
- Fix the time bug (display the local sunset/sunrise and current times)

JennieDalgren (approved):
⭐

PR #417, opened 2024-09-29
Weather app - Nella

nella-x:
Netlify link:
https://grayslay.netlify.app/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,11 @@
 # Weather App
 
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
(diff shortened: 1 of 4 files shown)


HIPPIEKICK (approved):
**HTML/CSS**
- Simple and clean HTML file. Consider using semantic elements more and keep the amount of divs down
- I love the quirky design (haha and the name of your page)! However, remember to try to follow our design the next time we tell you to, as it's an important skill to learn.

**JavaScript**
- Nice to gather all the icons in a variable! Since it's a list however, maybe it would make more sense as an array?
- Your code definitely works like it is now, but going forward I would like you to modularize your code more and break it out into functions.

**Clean Code**
- Clean up all console.logs, unused code and instructional comments from us.

Keep up the good work Nella!

PR #418, opened 2024-09-29
Annas Weather App

Anna2024WebDev:
## Netlify link
https://checkthe-weather.netlify.app/


index.html
```diff
@@ -0,0 +1,48 @@
...
+</html>
+
+
+
+
+
+    
+    
+  
+
```
HIPPIEKICK: Remove all empty lines

HIPPIEKICK (dismissed):
**HTML/CSS**
- 
- Very nice job following the design! Consider adding a fixed width to your `li` items to create some space in between
- 

**JavaScript**
- Really good job structuring your JavaScript! You've used clear and descriptive names on most of your variables and functions
- Good idea using toLowerCase and includes for the weather conditions 👍 

**Clean Code**
- The indentation is off in your HTML file. Also, I'd suggest you change to 2 spaces indentation in your JS file as it becomes hard to read when the indentations are this big.
- Don't mix arrow functions with the function keyword
- Since city is not a constant, it should be named in camelCase

**Changes Requested**
- Clean up the code as requested ☝️ Apart from that - good job!

index.html
Anna2024WebDev: Thank you for your feedback @HIPPIEKICK! I think I have updated according to the changes requested! Best, Anna

JennieDalgren (approved):
⭐️

PR #423, opened 2024-09-29
Joyce Kuo Weather App

JoyceKuode:
https://joycekuoweatherapp.netlify.app/


HIPPIEKICK (approved):
**HTML/CSS**
- Well structured HTML file! Just remember to put the script within the body. Also, consider using more semantic elements instead of just divs
- Real nice work with copying the design! Love the color overlay!
- Would be super nice if the arrow to change city had a pointer cursor on hover
- Try to keep the JS to the JS file (using event listeners instead of the `onclick` method)

**JavaScript**
- Well structured and easy-to-follow JS 👍 A lot of well-named and well-structured functions. The capitalizeFirstLetter function is beautiful!

**Clean Code**
- Be consistent with if you're using single or double quotes

No changes requested, just keep up the good work! 🥳

PR #425, opened 2024-09-29
project-7-weather-app

erikamolsson:
Netlify link: https://project-7-weather-app.netlify.app/

### DIFF
--- a/code/script.js
+++ b/code/script.js
@@ -2,7 +2,7 @@
 /* JSON file - Weather API */
 const API_KEY = "3cc790b1653416aade4ba256c95e0c67"
 const BASE_URL = "https://api.openweathermap.org/data/2.5/weather"
-let cityWeather = "Berlin";
+let cityWeather = "katrineholm";
 
 const URL = `${BASE_URL}?q=${cityWeather}&units=metric&APPID=${API_KEY}`
 console.log(URL)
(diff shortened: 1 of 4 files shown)


JennieDalgren (changes requested):
Great job with this project!

You are on the right track and have implemented the first features in a good way.
The searchbar is a stretch goal so I suggest not to focus on that at the moment.
The only thing that is missing to meet the requirements is the **weather forecast**

Try to follow that section in the instructions:

> Feature: Weather forecast 📅
> Show a forecast for the next 4 days. You can choose how to display the forecast - perhaps you want to show the min and max temperature for each day, or perhaps you want to show the temperature from the middle of the day, or the humidity, what it feels like and so on. Just make sure to make it all fit nicely with your chosen design.
> 
> https://api.openweathermap.org/data/2.5/forecast?q=Stockholm,Sweden&units=metric&APPID=YOUR_API_KEY
> The API gives us the next 4-5 days but for every third hour. So a good idea could be to only use the weather data from the same time every day. You can filter the forecast list array to only get the info from 12:00 each day for example.
> 
> Read the [endpoint documentation](https://openweathermap.org/forecast5) for the forecast.

Also don't forget to check previous questions and solutions to this on Stack Overflow. Many students have been struggling with this in previos cohorts too.
https://stackoverflowteams.com/c/technigo/questions/tagged/55

HIPPIEKICK (changes requested):
Sevilla weather looks good! Please remove the search bar until fully implemented so that it doesn't appear buggy.

code/script.js
```diff
@@ -0,0 +1,174 @@
...
+         // Prevent the default action of the Enter key (form submission, etc.)
+         event.preventDefault();
+
+         // Get the value from the search input field
+         cityWeather = searchInput.value;
+
+         // Log the search query
+         console.log('Search query (Enter pressed):', cityWeather);
+     }
+ });
```
HIPPIEKICK: I think you need to comment this out as well, because right now it's broken 👀

HIPPIEKICK (dismissed):

JennieDalgren (approved):
Approved ⭐

PR #427, opened 2024-09-29
Weather-app

zoe-lindqvist:
## Netlify link
https://weather-free-api.netlify.app/


JennieDalgren (approved):
Great job with this project!
I really like the extra effort with the geolocation. So nice!!!
You've met the requirements.

One tiny thing to change when you have time, to make the project not look broken, on small screen the search bar gets placed on top on the temperature.
image

Really good job with usage of include() method. Just be aware that if you have data that would be called for example *smash* then this would give you true `weather.includes("ash")` because ash is included in smash. So it is really important to understand the data you are working with. for example if the value will be exactly "ash" "haze" "fog" etc you can also do the checks like:
```
if (weather === "ash") etc...
```
Just a clarfication.

Another tiny UX detail, if you wrap your input and button for the search bar into a form element and put the submit event on the form instead, the user can use both the button or the enter key to search. 

Good structure of the code and good use of comments.
Nice job with error handling too.

PR #430, opened 2024-09-30
WeathertheWeatherApp

Fannyhenriques:
## Netlify link
https://master--weathertheweatherapp.netlify.app


JennieDalgren (approved):
Good job with this project!

Nice use of some helper functions and adding comments to the code.
It looks like the next step would be to refactor and see if you can find a way to keep your code a bit cleaner using the DRY approach. This is something you will learn with time, just something to reflect on. 

Also, avoid naming a variable as specific as you've done with `stockholm` and `stockholmTemp`. Even though this project only displays Stockholm weather, try to keep it more generic to make sense if you choose to swap to Tokyo for example. 
`city` and `cityTemp` might be better names.

Keep it up 💪

PR #431, opened 2024-09-30
Weather app - Jenny A

jempa182:
https://whats-the-weather-jenny.netlify.app/


code/script.js
```diff
@@ -0,0 +1,180 @@
...
+      return item.date.getDate() !== today && hour >= 6 && hour <= 18;
+    })
+    // Reduce to max 5 items, one per day
+    .reduce((acc, item) => {
+      if (acc.length < 5 && (acc.length === 0 || item.date.getDate() !== acc[acc.length - 1].date.getDate())) {
+        acc.push(item);
+      }
+      return acc;
+    }, []);
+};
```
JennieDalgren, line 134: Good way to use these methods. It is a bit advanced to do all this in one go like you are doing. I believe youi might have gotten help with structuring this code, which is totally fine. However it is superimportant that you really understand the code. So good that you have added comments to it, but also try to find the opportunity to actually explain this code to someone. Either in your team, in a short screen recording to the class, to a friend or family member (even though they might not understand coding) This will help you befästa kunskapen and also prepare you for upcoming technical interviews and demos "in real world"

instructions.md
```diff
@@ -115,3 +115,4 @@ Explore the API and use another endpoint of the Weather API to include supplemen
 
 **Feature: CSS Animations**  
 Add some CSS animations to your app, e.g. pulsating sun/raindrops.
+okoi
```
JennieDalgren, line 118: ? 😉

JennieDalgren (approved):
Great job with this project. 

Good use of comments the explain your code
Good use of functions, really impressed by the way you are splitting up the code in understandable and perfect sized functions. Good naming as well.

I made some comments in the lines of code too regarding adding code that you might have gotten from a mentor, friend, other code, AI etc.

I suggest that you set up your indentations for you projects to 2 spaces, it makes it easier to read (html file specifically.

You don't need to change anything to get approved! These are just suggestions going forward.

Keep up the good work 💪

PR #432, opened 2024-10-04
Xings Project weather a11y -  this is a hand-in for the Accessibility week.

xingyin2024:
https://xingsweatherapp.netlify.app/


HIPPIEKICK (approved):
Nice job with the accessibility work! The aria label on the search icon makes total sense, and good that you have a label for the search input as well. Your usage of semantic elements also ensures accessibility. The lighthouse report came back good and your contrast seems sufficient.

Just a note about tabindex, it’s usually used to make interactive elements navigable with keyboard. Since the forecast data is not clickable in anyway, I don’t think you need tabindex for that.

PR #436, opened 2024-10-06
Accessibility project (Weather app)

jempa182:
https://project-weather-a11y--whats-the-weather-jenny.netlify.app/


HIPPIEKICK (approved):
Good job with the accessibility work! Nice that you’ve implemented some aria labels 🥳 A note about tabindex however, it’s used for interactive elements, but since none of your elements are interactive/clickable, you don’t need tabindex for those. It’s more for, let’s say a hamburger menu that’s in reality a div - for the user to be able to "click it” using keyboard.

Good also that you’ve added alt tags for the icons, I think it’s easy to forget alt tags when you have the HTML in the JS 😅 Keep up the good work!

PR #438, opened 2024-12-19
Maryyy_ux_Weather-app

Maryyy-ux:
## Netlify link
(https://cute-selkie-735cdb.netlify.app/)


HIPPIEKICK (changes requested):
Have another look at the requirement and make sure you display the data in the expected format (e.g. weekdays). Something's buggy with the temperatures as well: temp, min and max are the same numbers.

Also - to make it look less buggy when starting off, give the city a starting value when declaring it so that you can fetch weather from somewhere.

